PR #17, opened 2019-09-18
Fix cross compilation on macOS

trha:
This is needed for cross compiling to `musl` on macOS.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,5 +1,5 @@
 {
-  nixpkgsPath ? <nixpkgs>,
+  nixpkgsPath ? ./nixpkgs.nix,
   system ? builtins.currentSystem,
   overlays ? [],
   crossSystem ? (import <nixpkgs/lib>).systems.examples.musl64,
--- a/overlay/real-host-triple.nix
+++ b/overlay/real-host-triple.nix
@@ -5,7 +5,7 @@ let
   kernel = platform.parsed.kernel.name;
   abi = platform.parsed.abi.name;
 in
-if platform.isWasi then
+if platform.isWasi or false then
   "${platform.parsed.cpu.name}-wasi"
 else {
   "i686-linux"        = "i686-unknown-linux-${abi}";
--- a/nixpkgs.nix
+++ b/nixpkgs.nix
@@ -7,7 +7,7 @@ import
   (args // {
     config.packageOverrides = pkgs: {
       openssl = (pkgs.openssl.override {
-        # `perl` is only used at build time, but the derivation incorrectly uses host perl
+        # `perl` is only used at build time, but the derivation incorrectly uses host `perl`
         # as an input.
         perl = pkgs.buildPackages.buildPackages.perl;
       }).overrideAttrs (_: {
@@ -16,10 +16,9 @@ import
         outputs = [ "bin" "dev" "out" ];
       });
 
-      # `libuv` needs rebuilding because of the harmless `openssl` change above.
-      # The test suite fails for some reason so just disable it.
+      # These packages need rebuilding because of the harmless `openssl` change above.
+      # The test suites for them fails (for some reason) so just disable them.
       libuv = pkgs.libuv.overrideAttrs (_: { doCheck = false; });
-
       e2fsprogs = pkgs.e2fsprogs.overrideAttrs (_: { doCheck = false; });
     };
   })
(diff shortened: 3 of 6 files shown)


ebkalderon:
Is this supposed to fix the `musl64` compilation of `cargo2nix` on macOS? If so, this pull request doesn't fix the cross compilation issue for me. I can confirm with `nix-docker` that this change doesn't affect the Linux build, which is good, but the macOS build, triggered with the following command:

```
# Notice there is no `--arg crossSystem null` here.
nix-build -A package
```

still fails with the following error:

```
builder for '/nix/store/8kqvc6wrz8hjyxan3kgll0kmnr6gdgwn-perl-5.30.0-x86_64-unknown-linux-musl.drv' failed with exit code 1
cannot build derivation '/nix/store/kkm6phjd2r91d1fygfvbq3vabs8qia31-openssl-1.1.1c-x86_64-unknown-linux-musl.drv': 1 dependencies couldn't be built
cannot build derivation '/nix/store/zg486an3wnlaq4qcc7nk2439p7mv4fka-nghttp2-1.39.2-x86_64-unknown-linux-musl.drv': 1 dependencies couldn't be built
cannot build derivation '/nix/store/rhbxphxihg60pyn1kddr9qdvbxbc29m1-openssl.drv': 1 dependencies couldn't be built
cannot build derivation '/nix/store/lmfv1nmvx5hnpgj5lz13smagyq47zkwy-crate-curl-sys-0.4.20-x86_64-unknown-linux-musl.drv': 1 dependencies couldn't be built
cannot build derivation '/nix/store/78x23d8rmk9fjrp8189dfa0859ni9bcs-crate-openssl-sys-0.9.49-x86_64-unknown-linux-musl.drv': 1 dependencies couldn't be built
cannot build derivation '/nix/store/bzdv1hsxhlbgybf4mvbkax5389gxp6wi-crate-cargo-0.37.0-x86_64-unknown-linux-musl.drv': 1 dependencies couldn't be built
cannot build derivation '/nix/store/pv8lml09iaq03h5d0fb58q3730fx2lcl-crate-cargo2nix-0.3.0-x86_64-unknown-linux-musl.drv': 1 dependencies couldn't be built
error: build of '/nix/store/pv8lml09iaq03h5d0fb58q3730fx2lcl-crate-cargo2nix-0.3.0-x86_64-unknown-linux-musl.drv' failed
```

I'm guessing that this change is purely to help facilitate the Mothership services' conversion process, correct?

trha:
@ebkalderon I've only tried cross compiling with `common-io`, not `cargo2nix` itself. I'll keep this PR open and look into `cargo2nix`today.

ebkalderon:
@trha That's great to hear that it works for `common-io`, though! Since it doesn't appear to affect the native Linux builds in any significant way, perhaps this PR would still have value to merge on its own, since it strictly enables more packages to be compiled without breaking existing functionality. Still, I guess we can leave it open if you want to investigate a bit more. Feel free to make any further changes you find, or perhaps ping me if you would like it to be merged as-is.

overlay/real-host-triple.nix
```diff
@@ -5,7 +5,7 @@ let
   kernel = platform.parsed.kernel.name;
   abi = platform.parsed.abi.name;
 in
-if platform.isWasi then
+if platform.isWasi or false then
```
dingxiangfei2009: Nice catch!
As a note, the `isWasi` symbol is available in upstream `nixpkgs` since June.

default.nix
```diff
@@ -1,5 +1,5 @@
 {
-  nixpkgsPath ? <nixpkgs>,
+  nixpkgsPath ? ./nixpkgs.nix,
```
dingxiangfei2009: Sure, let's pin the default `nixpkgs`.

overlay/real-host-triple.nix
ebkalderon: This is a huge help! I've been having issues building `cargo2nix` projects in NixOS environments using very old Nixpkgs (e.g. `nix-docker`) specifically because of this attribute not being present.

ebkalderon:
@trha Seems to have fixed the issue for me! 🎉 I can now cross-compile `cargo2nix` to `aarch64` on macOS, and I can also build a few other projects pinned to an older Nixpkgs that I couldn't build with `cargo2nix` earlier.

PR #20, opened 2019-10-09
Replace broken .envrc with a working shell

workflow:
Adds a working `direnv` config for hacking on cargo2nix itself


workflow:
> Very handy! Thanks, @workflow. Is there some way you can have the `shell.nix` pull in the `shell` attribute from the `default.nix` instead of calling `mkShell` yourself? That attribute should contain all the things necessary to iterate on `cargo2nix` locally, including Rust, pinned system dependencies, etc.
> 
> If there doesn't seem to be a way, then perhaps we can approve this PR as is.

Certainly can :)

PR #25, opened 2019-10-24
Document naming convention for the config format

workflow:
cc: @dingxiangfei2009 @antigravityla @trha


workflow:
There is a proposed change as well for this in #27, but I'll merge this one anyway since it won't cause any harm.

PR #26, opened 2019-10-24
Rename resolver -> srcFetcher

workflow:
Just a suggestion from @trha, not sure yet if this is the perfect name -)

cc: @dingxiangfei2009 @antigravityla


workflow:
Yeah, I also think this is superseded by #27

Issue #19, opened 2019-10-08
Provide more default values in overlay


ebkalderon:
Considering the high complexity of the `default.nix` referenced by most projects, it would be nice if we provided more default values in the `cargo2nix` overlay to reduce the initial boilerplate. Some examples include:

1. Restoring `<nixpkgs>` as the default, if left unspecified. This [used to be the default value](https://github.com/tenx-tech/cargo2nix/pull/17/files#diff-5712e736e0de6ba170577f8472c398e9R2), but was changed to help cross-compilation of `cargo2nix` specifically. The `nixpkgs.nix` isn't needed for projects using `cargo2nix` as a build system, so maybe the generated `default.nix` would simply use `<nixpkgs>` as the default (see #18).
2. Importing the `cargo2nix` overlay could also re-export the `nixpkgs-mozilla` overlay, with the latest Rust stable selected as the default and `aarch64-unknown-linux-gnu` enabled, if left unspecified.
3. Common patched dependencies such as `openssl` and `macosFrameworks` could be re-exported from the overlay, if needed.
4. `resolveResponse` and `rustPackagesWithResolve` could have a default value for simple cases where manual resolution isn't necessary or impractical for a given project.

Any thoughts, @dingxiangfei2009?

dingxiangfei2009:
1. I think this is a fair point, especially when I as a user just want to use whichever `<nixpkgs>` specified by the system.

2. The Mozilla overlay can be (?) specified as an additional input to the `cargo2nix` overlay function. Adding `aarch64-unknown-linux-gnu` is very welcomed, since this will highlight that this overlay is capable of handling cross-compilation.

3. Very welcomed! Tired of specifying it in every projects I am working on that has remotely depended on cryptography

4. Yes please. In fact, the expression in `default.nix` is canonical? I cannot draw useful variants or abstraction so far from my experience.

workflow:
@ebkalderon How would you assess the current state?

ebkalderon:
The most recent version of `overlay/overrides.nix` now provides a very comprehensive list of default overrides for packages, and all of the bullet points in the issue description are either now included in that file or are no longer applicable, such as `cargo2nix resolve` and `resolveResponse` which don't exist anymore. The `default.nix` has been whittled down to such a lightweight and small size that I believe this issue can be safely closed.

PR #27, opened 2019-10-29
Rework code generation

### LINKED ISSUE #21: Reduce dependency resolution time

trha:
Closes #21.


default.nix
```diff
@@ -1,157 +1,92 @@
...
-    then
-      with pkgs.darwin.apple_sdk.frameworks;
-      [Security CoreServices]
-    else [];
-
-  # choice of rustc
-  rustChannel = pkgs.buildPackages.rustChannelOf {
-    channel = "1.37.0";
+  rustChannel = buildPackages.rustChannelOf {
+    channel = "1.38.0";
```
workflow: I would bump the Rust version in a separate PR, unless there's a specific reason?

default.nix
```diff
@@ -1,157 +1,92 @@
...
   };
-
   inherit (rustChannel) cargo;
   rustc = rustChannel.rust.override {
     targets = [
       (pkgs.rustBuilder.rustLib.realHostTriple pkgs.stdenv.targetPlatform)
-      "aarch64-unknown-linux-gnu"
     ];
+    extensions = [ "rust-std" ];
```
workflow: Isn't that provided by default?

default.nix
```diff
@@ -1,157 +1,92 @@
...
+              envBuildPlatform = envize pkgs.buildPlatform.config;
+              envHostPlatform = envize pkgs.hostPlatform.config;
+              staticOpenssl = pkgs: (pkgs.openssl.override {
+                # `perl` is only used at build time, but the derivation incorrectly uses host `perl` as an input.
+                perl = pkgs.buildPackages.buildPackages.perl;
+                static = true;
+              }).overrideAttrs (drv: {
+                installTargets = "install_sw";
+                outputs = [ "dev" "out" "bin" ];
+                postInstall = builtins.replaceStrings ["rm -r " "rmdir "] ["rm -rf " "rm -rf "] drv.postInstall;
```
workflow: :thinking:

default.nix
```diff
@@ -1,157 +1,92 @@
...
-    inherit (rustPackages.config) features;
+  rustPkgs = pkgs.rustBuilder.makePackageSet {
+    inherit cargo rustc packageFun rustPackageConfig;
+    buildRustPackages = buildPackages.rustBuilder.makePackageSet {
+      inherit cargo rustc packageFun rustPackageConfig;
+    };
   };
-}
+in
+  rustPkgs."cargo2nix2 0.1.0 unknown" { }
```
workflow: I like that, and was thinking whether we can come up with some shorter aliases here. Like

`unknown` -> `local`
`registry+https://github.com/rust-lang/crates.io-index` -> `crates.io`

Cargo.nix
```diff
@@ -0,0 +1,3487 @@
...
+    };
+    features = builtins.concatLists [
+    ];
+    dependencies = {
+    };
+    devDependencies = {
+    };
+    buildDependencies = {
+    };
+    manifest = builtins.fromTOML "[package]\nauthors = [\"Remi Rampin <remirampin@gmail.com>\"]\ndescription = \"Minimal Adler32 implementation for Rust.\"\ndocumentation = \"https://remram44.github.io/adler32-rs/index.html\"\nkeywords = [\"adler32\", \"hash\", \"rolling\"]\nlicense = \"Zlib\"\nname = \"adler32\"\nreadme = \"README.md\"\nrepository = \"https://github.com/remram44/adler32-rs\"\nversion = \"1.0.4\"\n";
```
ebkalderon: I'm curious: do we even use these attributes for anything in any stage of `cargo2nix`? I know the original `Cargo.lock` doesn't even include these fields at all, so perhaps we can omit this key altogether?

default.nix
```diff
@@ -1,157 +1,92 @@
...
+          "registry+https://github.com/rust-lang/crates.io-index".curl-sys."*" = darwinFrameworks;
+          "registry+https://github.com/rust-lang/crates.io-index".libgit2-sys."*" = darwinFrameworks;
+        };
+        buildInputs = {
+          "registry+https://github.com/rust-lang/crates.io-index".libgit2-sys."*" = [ pkgs.libiconv ];
+          "registry+https://github.com/rust-lang/crates.io-index".cargo."*" = [ pkgs.libiconv ] ++ darwinFrameworks;
+          unknown.cargo2nix2."*" = [ pkgs.libiconv ] ++ darwinFrameworks;
+        };
+        environment = {
+          "registry+https://github.com/rust-lang/crates.io-index".openssl-sys."*" =
```
ebkalderon: This entire let-in block seems quite complex to include in the `default.nix`. Could this maybe be added to the overlay and re-exported so it can be re-used, as suggested in #19?

default.nix
ebkalderon: It seems that this pull request completely removes the `shell` feature, the automatic vendoring into the Nix store with `.cargo/config` source replacement, and also airplane mode. Do you intend to put this functionality back in?

ebkalderon: I think the latter value is pretty important to keep as-is, though, since that name isn't just an alias; it corresponds directly to the registry values specified in the `Cargo.lock`. Changing the format of these URIs would break behavior with upstream Cargo. Plus, it would be pretty inconsistent to use a value here that differs from the one used in the `Cargo.lock` and also the `buildInputs`/`environment` section above.

I wouldn't mind changing `unknown` to `local`, though, since it's just a placeholder, although it could potentially be confused for the `local-registry` source ID already provided by Cargo.

workflow: Maybe `none`?

workflow: Or `noregistry`

ebkalderon: Sure! Either of those could work too. ☺️

overlay/lib/fetch.nix
```diff
@@ -0,0 +1,31 @@
...
+        name = "${name}-${version}-src";
+        inherit url rev sha256;
+      };
+    in
+      /. + builtins.readFile (runCommand "find-crate-${name}-${version}"
+        { nativeBuildInputs = [ jq remarshal ]; }
+        ''
+          shopt -s globstar
+          for f in ${repo}/**/Cargo.toml; do
+            if [[ "$(remarshal -if toml -of json "$f" | jq '.package.name == "${name}" and .package.version == "${version}"')" == "true" ]]; then
```
edude03: Doesn't nix have native toml parsing? You should be able to use that instead. Unless I'm misunderstanding what this does 😓

default.nix
trha: Yeah, as we're not actively using this functionality, I plan to bring it back later.

overlay/lib/fetch.nix
trha: It does, but this is easier to do in bash, as there's no way to recursively traverse a directory in pure Nix.

overlay/make-package-set.nix
```diff
@@ -9,14 +9,16 @@
   rustc,
   mkRustCrate,
   buildRustPackages ? null,
-  target ? null,
+  localPatterns ? [ ''^(src)(/.*)?'' ''[^/]*\.(rs|toml)$'' ],
```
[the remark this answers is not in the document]
trha: Yeah, but it's overridable. The default value only covers the minimum.

Cargo.nix
trha: We do. `cargo` needs the manifest to provide [these environment variables](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates) during compilation.

default.nix
trha: Good idea! I'll leave this to a later PR where I plan to introduce a more composable way to provide config.

overlay/make-package-set.nix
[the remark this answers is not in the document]
ebkalderon: Ah, that makes sense. We should probably document this somewhere, since it's not immediately obvious from the README that this option exists.

Cargo.nix
ebkalderon: If the crate source itself is being vendored in the Nix store via `fetchCratesIo`, presumably including the `Cargo.toml`, wouldn't you be able to read the contents of the file from disk? Why does it need to be embedded directly into the `Cargo.nix`?

default.nix
ebkalderon: That's fair! But since this is a large breaking change, we should update the README to reflect this and open an issue to reintroduce the functionality later, if we are planning to merge this pull request directly into `master` without adding it back.

default.nix
ebkalderon: Sounds good! 👌

Cargo.nix
trha: The source we get from `fetchCratesIo` is a tarball, which won't be unpacked until the build runs. We need the manifest before the build to prepare a patched manifest. We can also do this in bash during the build, but it's a bit more complicated. I'll look into this.

ebkalderon: Got it! I see why it would be simpler to have it embedded in the `Cargo.nix` at generation time, but then again, it would be great to get rid of the extra "fluff" in the generated Nix code. It's definitely not a priority to remove it, but it sure would be cool to do so down the line. 😄

default.nix
trha: I'll leave this to a later PR.

trha:
This PR is now ready for review. I've updated `default.nix` to include detailed instructions on how to bootstrap `cargo2nix`.

default.nix
```diff
@@ -1,159 +1,114 @@
...
+    packageFun = import ./Cargo.nix;
   };
-}
+in
+  # `rustPkgs` now contains all crates in the dependency graph.
+  # To build normal binaries, use `rustPkgs.<registry>.<crate>.<version> { }`.
+  # To build test binaries (equivalent to `cargo build --tests`), use
+  #   `rustPkgs.<registry>.<crate>.<version> { complieMode = "test"; }`.
+  # To build bench binaries (equivalent to `cargo build --benches`), use
+  #   `rustPkgs.<registry>.<crate>.<version> { complieMode = "bench"; }`.
```
ebkalderon: A few tiny spelling mistakes here (`complieMode` instead of `compileMode`).

trha:
Sorry for the repeated review requests. I thought this PR were ready, but then found many blocking issues. It's ready for real now.

default.nix
```diff
@@ -1,159 +1,116 @@
...
+  # - `rustPackageConfig` (optional): See above.
+  # - `localPatterns` (optional):
+  #     A list of regular expressions that specify what should be included in the sources of your workspace's crates.
+  #     The expressions are relative to each crate's manifest directory.
+  #     This argument is optional and defaults to include the `src` directory and all `toml` files at the root of the manifest directory.
+  # - `rootFeatures` (optional):
+  #     A list of activated features on your workspace's crates.
+  #     Each feature should be of the form `<crate_name>[/<feature>]`.
+  #     If `/<feature>` is omitted, the crate is activated with no default features.
+  # - `release` (optional): Whether to enable release mode (equivalent to `cargo build --release`), defaults to `true`.
```
workflow: :heart:

default.nix
```diff
@@ -1,159 +1,116 @@
...
+    rustChannel = "1.37";
+    packageFun = import ./Cargo.nix;
   };
+in
+  # `rustPkgs` now contains all crates in the dependency graph.
+  # To build normal binaries, use `rustPkgs.<registry>.<crate>.<version>`.
+  # To build test binaries (equivalent to `cargo build --tests`), use
+  #   `rustPkgs.<registry>.<crate>.<version>.override { compileMode = "test"; }`.
+  # To build bench binaries (equivalent to `cargo build --benches`), use
+  #   `rustPkgs.<registry>.<crate>.<version>.override { compileMode = "bench"; }`.
```
workflow: :heart: :heart: :heart:

overlay/lib/fetch.nix
```diff
@@ -1,9 +1,21 @@
 { buildPackages, lib }:
+let
+  unpackSrc = { name, src }: buildPackages.stdenvNoCC.mkDerivation {
+    inherit name src;
+    preferLocalBuild = true;
+    allowSubstitutes = false;
```
workflow: Technically the `unpackPhase` could happen on a remote builder (I think) ;)

overlay/default.nix
```diff
@@ -9,7 +10,9 @@ let
 
     rustLib = callPackage ./lib { };
 
-    makePackageSet = callPackage ./make-package-set.nix;
+    makePackageSet = callPackage ./make-package-set/full.nix;
+
+    makePackageSet' = pkgs.callPackage ./make-package-set/simplified.nix;
```
workflow: Curious: What's the simplified one used for?

trha: It removes a lot of boilerplate (like choosing `rustc`, `cargo`, specifying build packages, ...).

overlay/lib/fetch.nix
trha: > `allowSubstitutes`
> 
> If this attribute is set to false, then Nix will always build this derivation; it will not try to substitute its outputs. This is useful for very trivial derivations (such as writeText in Nixpkgs) that are cheaper to build than to substitute from a binary cache.

It's usually faster to just unpack locally instead of downloading from a cache so I set this to `false`.

PR #29, opened 2019-11-05
Propagate profiles

trha:
This PR adds support for the [`[profile.*]`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-profile-sections) sections defined at the top level `Cargo.toml`.

### DIFF
--- a/Cargo.nix
+++ b/Cargo.nix
@@ -1,7 +1,3 @@
-let
-  profilesByName = {
-  };
-in
 {
   release ? true,
   rootFeatures ? [
@@ -16,6 +12,8 @@ in
 }:
 let
   inherit (rustLib) fetchCratesIo fetchCrateLocal fetchCrateGit expandFeatures decideProfile genDrvsByProfile;
+  profilesByName = {
+  };
   rootFeatures' = expandFeatures rootFeatures;
   overridableMkRustCrate = f:
     let
--- a/Cargo.nix
+++ b/Cargo.nix
@@ -1,3 +1,7 @@
+let
+  profilesByName = {
+  };
+in
 {
   release ? true,
   rootFeatures ? [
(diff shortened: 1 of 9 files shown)


Cargo.nix
```diff
@@ -1,22 +1,31 @@
+let
+  profilesByName = {
+  };
+in
```
ebkalderon: Minor nitpick, but does this need to be defined in a separate let-in block? I think it could also go in the let-in block beginning on lines 17-18, and it would look a tad nicer.

overlay/lib/profiles.nix
```diff
@@ -0,0 +1,37 @@
+let
+  profileNames = [
+    "release"
+    "dev"
+    "test"
+    "bench"
+    "__noProfile" # A placeholder profile for build scripts, which don't respect any profiles.
```
ebkalderon: I understand that `__noProfile` is used only for build scripts, but I believe those are always built in `"dev"` mode. If so, why is this preferable over embedding `profile = "dev"` in the `Cargo.nix` at generation time?

trha: Build scripts only inherit (`opt-level` and `debug`)[https://github.com/rust-lang/cargo/blob/rust-1.38.0/src/cargo/core/profiles.rs#L138] from the current profile. I didn't take this into account to simplify things.

Cargo.nix
trha: I move this outside of the returned function to avoid allocating the profiles every time that function is called. But on second thought, that function is only called twice, so yeah, I'll just move it back.

overlay/lib/profiles.nix
ebkalderon: If they don't make an appreciable difference, could we not select the `"dev"` profile for build scripts in that case?

trha: 😞 User can set `panic = "abort"` in `[profile.dev]`, which unfortunately isn't compatible with build scripts (which include `build.rs` and proc macros).

ebkalderon: Ah, makes sense. 😞

PR #30, opened 2019-11-07
Overrides

trha:
This PR replaces `config` with `overrides`, a more reusable and composable mechanism to provide native inputs to Rust crates.

### DIFF
--- a/overlay/lib/overrides.nix
+++ b/overlay/lib/overrides.nix
@@ -1,5 +1,5 @@
 # An overrider has the ability to modify the arguments passed to a function. It takes the original set of arguments
-# and returns a new set of arguments, which will then be merged will the original set to become the new arguments
+# and returns a new set of arguments, which will then be merged with the original set to become the new arguments
 # to the original function.
 # type Overrider = Attrs -> Attrs
 #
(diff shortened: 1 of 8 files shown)


overlay/lib/overrides.nix
```diff
@@ -0,0 +1,57 @@
+# An overrider has the ability to modify the arguments passed to a function. It takes the original set of arguments
+# and returns a new set of arguments, which will then be merged will the original set to become the new arguments
```
workflow:
```suggestion
# and returns a new set of arguments, which will then be merged with the original set to become the new arguments
```

overlay/overrides.nix
```diff
@@ -0,0 +1,120 @@
...
+      nghttp2 = pkgs.nghttp2.override { inherit openssl; };
+      libssh2 = pkgs.libssh2.override { inherit openssl; };
+      libkrb5 = pkgs.libkrb5.override { inherit openssl; };
+    };
+
+in
+rec {
+  patches = [ patchOpenssl patchCurl ];
+
+  # Don't forget to add your new overrides here.
```
workflow: Might make sense to separate this out into an `all-overrides.nix` file a la Nixpkgs `all-packages.nix` style

trha: Overrides are much smaller than packages so I think we can put everything in one file for now.

PR #37, opened 2019-11-20
Propagate environment variables to root crates

### LINKED ISSUE #33: Direnv failing
...because of missing `shell.nix`

### LINKED ISSUE #34: Shell instructions broken
Current README instructions for the "Optional declarative development shell" are broken

trha:
This PR allows overrides to propagate environment variables further up the dependency graph, which reduces the work to create a development shell.

CC @edude03 @antigravityla.

Resolves #33 
Resolves #34

### DIFF
--- a/overlay/mkcrate-nobuild.nix
+++ b/overlay/mkcrate-nobuild.nix
@@ -1,0 +1,37 @@
+# Creates a derivation for a crate for inputs propagation only.
+{ lib, stdenv }:
+{
+  release, # Compiling in release mode?
+  name,
+  version,
+  registry,
+  src,
+  features,
+  dependencies,
+  devDependencies,
+  buildDependencies,
+  compileMode ? "build",
+  profile,
+  meta ? { },
+  rustcflags ? [ ],
+  rustcBuildFlags ? [ ],
+}:
+with lib; with builtins;
+let
+  inherit
+    (({ right, wrong }: { runtimeDependencies = right; buildtimeDependencies = wrong; })
+      (partition (drv: drv.stdenv.hostPlatform == stdenv.hostPlatform)
+        (concatLists [
+          (attrValues dependencies)
+          (optionals (compileMode == "test") (attrValues devDependencies))
+          (attrValues buildDependencies)
+        ])))
+    runtimeDependencies buildtimeDependencies;
+in stdenv.mkDerivation {
+  name = "crate-${name}-${version}";
+  buildInputs = runtimeDependencies;
+  propagatedBuildInputs = concatMap (drv: drv.propagatedBuildInputs) runtimeDependencies;
+  nativeBuildInputs = buildtimeDependencies;
+  phases = "installPhase fixupPhase";
+  installPhase = "mkdir -p $out";
+}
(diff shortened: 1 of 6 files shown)


edude03:
Can you explain how this works / what its supposed to do? I'm having a hard time come up with a mental image of what the code does when operating.

overlay/overrides.nix
```diff
@@ -4,7 +4,7 @@ let
   envize = s: builtins.replaceStrings ["-"] ["_"] (lib.toUpper s);
 
   patchOpenssl = pkgs: (pkgs.openssl.override {
-    # We only need `perl` at build time. It's also used as the interpreter for one
+    # We only need `perl` at build time. x It's also used as the interpreter for one
```
antigravityla: nit

overlay/overrides.nix
```diff
@@ -57,6 +57,8 @@ let
 in rec {
   patches = { inherit patchOpenssl patchCurl patchPostgresql joinOpenssl propagateEnv; };
 
+  a = pkgs.openssl;
```
antigravityla: nit

antigravityla:
@edude03 in the end it sets some environment variables so that vanilla cargo tooling will build with overridden dependencies.  

It can do this for any package that observes environment variables to pick up override paths.  Many sys crates do this.  

If we make a crate that we want to inject into the dev shell, we just record the output path in the appropriate variable.  If there is no variable, we can patch the crate, possibly in upstream.  it's not super nice.  It will work.

overlay/overrides.nix
```diff
@@ -96,20 +113,26 @@ in rec {
...
       };
 
   prost-build = makeOverride {
     name = "prost-build";
-    overrideAttrs = _: {
-      PROTOC = "${pkgs.buildPackages.buildPackages.protobuf}/bin/protoc";
+    overrideAttrs = drv: {
+      propagatedBuildInputs = drv.propagatedBuildInputs or [ ] ++ [
+        (propagateEnv [
+          { name = "PROTOC"; value = "${pkgs.buildPackages.buildPackages.protobuf}/bin/protoc"; }
```
antigravityla: @edude03   one such override that we propagate into the environment

overlay/mkcrate-nobuild.nix
```diff
@@ -0,0 +1,39 @@
...
+        ])))
+    runtimeDependencies buildtimeDependencies;
+in stdenv.mkDerivation {
+  name = "crate-${name}-${version}";
+  buildInputs = runtimeDependencies;
+  propagatedBuildInputs = concatMap (drv: drv.propagatedBuildInputs) runtimeDependencies;
+  nativeBuildInputs = buildtimeDependencies;
+  phases = "installPhase fixupPhase";
+  installPhase = "mkdir -p $out";
+  preferLocalBuild = true;
```
antigravityla: Yeah, I think farming the crate itself out should be... fine?  The only advantage to doing the rust build locally AFAIK is to populate `/target/` for vanilla cargo tools to re-use in local user workflows.

trha: This is a special version of `mkRustCrate` which does nothing but `mkdir -p $out` and creating Nix support files (which is exactly the same as what `propagateEnv` does).

antigravityla:
For installing system-wide, the command did change to `nix-env -if . -A package` though no idea if this was an intended use pattern.  Bootstrapping a project got a bit weirder with this.

PR #39, opened 2019-12-05
Support for alternative registries

trha:


overlay/lib/fetch.nix
```diff
@@ -40,4 +39,34 @@ in
           echo Crate ${name}-${version} not found in ${url}
           exit 1
         '');
+
+  # This implementation of `fetchCrateAlternativeRegistry` assumes that the download URL is updated frequently
+  # on the registry index as a workaround for the lack of authentication for crate downloads. Each time a crate needs
+  # to be downloaded, the whole index needs to be recloned to get the download URL, which is potentially expensive.
```
snawaz: > Each time a crate needs to be downloaded, the whole index needs to be recloned to get the download URL, which is potentially expensive.

Within a single build-run, we can safely assume that the download URL wont change though. Does the implementation take advantage of this?

trha: How does it get the URL in the first place?

snawaz: Well, what I meant is that.. when a build starts, it'll clone then index.. and during the build, if it needs to download multiple crates.. one after another.. it can use the same URL for all those downloads. Does that make sense? I'm not sure how it works in this case though.. but I imagine what I just described.

snawaz: And when I said, _build starts_, I meant the build of the project (the root build, not the dependencies).

trha: > it'll clone the index.. 

The index has to be each time a crate from that index needs downloading. We can't cache the index because we don't know its SHA beforehand.

snawaz: Let's chat on Slack. I didn't get it properly.

src/main.rs
```diff
@@ -626,6 +629,23 @@ fn write_source_nix<W: Write>(
...
+        writeln!(f, "{} {{", scope.fetch_crate_alternative_registry)?;
+        {
+            let mut f = f.indent(2);
+            writeln!(f, "index = {};", source_id.url())?;
+            writeln!(f, "name = {:?};", p.name())?;
+            writeln!(f, "version = {:?};", p.version().to_string())?;
+            writeln!(
+                f,
+                "sha256 = {:?};",
+                checksum.unwrap_or_else(|| panic!(
```
ebkalderon: Minor point, I think `unwrap_or_else(|| panic!(...))` is fully equivalent to `expect()`, so maybe we could just use that here?

trha: They're not exactly equivalent. `expect(&format(..))` allocates in all cases, where `unwrap_or_else` only allocates in the error case. 😅 This doesn't make a difference most of the time, just a habit I learned.

snawaz: It's not guaranteed &mdash; but if the compiler is smart enough, these two will emit similar machine code, given that `expect()` is a *branch*, and so is `unwrap_*`.

trha: `expect` is not a branch. Arguments passed to `expect` are always evaluated eagerly, as opposed to `unwrap_or_else`.

snawaz: Well, there has to be branch. `r.explicit(..)` is more or less same as `if let Some(x) = r { x } else { panic!(..) }`

So if we have these two functions:

```rust
pub fn expect_branch_implicit(r: Option<i32>) -> i32 {
    r.expect("test untest expect")
}
pub fn expect_branch_explicit(r: Option<i32>) -> i32 {
    if let Some(x) = r {
        x
    } else {
        panic!("test untest expect")
    }
}
```

The compiler at `-C opt-level=3` emits this (comments are mine) which is almost identical:

```assembly
; this is machine code for expect_branch_implicit

_ZN7example22expect_branch_implicit17h92ba2b7ec23e01c1E:
        push    rax
        test    edi, edi
        je      .LBB11_1
        mov     eax, esi
        pop     rcx
        ret
.LBB11_1:
        lea     rdi, [rip + .L__unnamed_4]
        mov     esi, 18
        call    qword ptr [rip + _ZN4core6option13expect_failed17h616db3795314a1d2E@GOTPCREL]
        ud2

# ------------------------------------------------------------
; and this is the machine code for expect_branch_explicit

_ZN7example22expect_branch_explicit17h5007129a9ce121b8E:
        push    rax
        cmp     edi, 1
        jne     .LBB12_1
        mov     eax, esi
        pop     rcx
        ret
.LBB12_1:
        lea     rdi, [rip + .L__unnamed_7]
        call    _ZN3std9panicking11begin_panic17h544461351da18525E
        ud2
```

You see both has branch? The only difference is that one checks for _equality_ (`test` followed by `je`) and other checks for _inequality_ (`cmp` followed by `jne`). The same is true of `unwrap` and `unwrap_*`.

---

Now coming back to your second point: 

> *Arguments passed to expect are always evaluated eagerly, as opposed to unwrap_or_else*.

That is only semantically required, from the language point of view. However, the compiler could optimize things in a way that may not execute things eagerly as you described it. In such optimizations, the compiler only _pretends_ as if it has evaluated things as required by the language semantics, and then later in the light of more information available at runtime, it sees whether it really needs to evaluate them or not at all... as long as its optimization strategy does not change the _observable behaviors_ of the program, it can do that. This is referred to as [as-if][as-if] rule. It's quite popular in C and C++. However, it applies to many languages (even the interpreted ones).

[as-if]:https://en.wikipedia.org/wiki/As-if_rule

snawaz: Also, if we use `match r` instead of `if let .. `, then branch is _identical_:

```

Comment (trha) [reply src/main.rs]: You're basically expecting the compiler to optimize this
```rust
let msg = format!(..);
if let Err(_) = result {
  panic!("{}", msg);
}
```
to
```rust

if let Err(_) = result {
  let msg = format!(..);
  panic!("{}", msg);
}
```

This optimization is safe only if the lowered expression is free from side effects. This is not the case with `format!`, as it allocates on the heap.

snawaz: 1. I dont claim that the compiler will do that. I merely said if the compiler is smart enough.. 
2. You'll be surprised to know that memory allocation is NOT considered side-effects .. the way you're saying it (_memory allocations_ are not _observable behaviors_ of the programs). The only side-affects that compilers are NOT allowed to eliminate is IO. Almost everything else belongs to the compiler whims!

trha: > memory allocation is NOT considered side-effects

Maybe in other languages, but not in Rust.

snawaz: I'm pretty sure that it is true of Rust as well. Rust uses LLVM for such things. So it inherits many things automatically. Also, Rust (currently) does not have _formal_ specification/model that defines such behaviors. It depends on the LLVM for such things.

PR #41, opened 2019-12-07
Include integration test binaries in test mode

### LINKED ISSUE #40: Integration tests are not included in output
The officially supported method of producing test binaries in a `cargo2nix` project is to build the crate with `compileMode = "test"`. Unfortunately, it appears that among the test binaries produced, any integration test binaries from the top-level `tests` directory are not included.

It seems that inline library tests, `main.rs` tests, and `src/bin/*` tests are included in the output, but integration tests from the `tests` directory are not. Like the other tests, the integration tests produced by Cargo are located in `target/<host-triple>/<profile>/<test-name>-<hash>`. I believe we should include these in the derivation output as well.

ebkalderon:
### Fixed

* Add top-level `tests` directory to `localPatterns` in `overlay/mkcrate.nix`.

As described in #40, `cargo2nix` currently does not output integration tests by default when building the crate in test mode. This is because `localPatterns` does not include the top-level `tests` directory in the default path regex. This directory is standard for any Cargo project, and the current behavior of not including integration tests when `compileMode = "test"` is quite surprising, especially for a new user. I believe it should be fine to include this directory by default, if it exists.

Fixes #40.

### DIFF
--- a/overlay/make-package-set/full.nix
+++ b/overlay/make-package-set/full.nix
@@ -11,7 +11,7 @@
   cargo,
   rustc,
   buildRustPackages ? null,
-  localPatterns ? [ ''^(src)(/.*)?'' ''[^/]*\.(rs|toml)$'' ],
+  localPatterns ? [ ''^(src|test)(/.*)?'' ''[^/]*\.(rs|toml)$'' ],
   packageOverrides ? [ ],
   fetchCrateAlternativeRegistry ? _: throw "fetchCrateAlternativeRegistry is required, but not specified in makePackageSet",
   release ? null,


ebkalderon:
Please advise whether this change warrants a new release.

ebkalderon:
I believe we should release 0.6.0 after this PR gets merged since both the default package overrides and alternate registry support features are not included in 0.5.0.

ebkalderon:
No worries, @trha! Changelogs are nice to have in any project, especially in open source projects. I feel they serve a crucial part of the semver stability contract between developer and user, describing succinctly what changes went into the version bump before they upgrade. I happen to be partial to the [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) standard, so I tried to base them on that.

Though, before I merge this PR, I've just realized that I have a minor grammatical mistake in my commit message. Might need to ask you to reapprove in a minute or two, haha.

ebkalderon:
@trha Okay, just fixed the commit message. Would you mind reapproving the PR, please?

PR #46, opened 2019-12-10
Generate a set of all workspace crates

trha:
Becase `unknown.<crate>.<version>` is just too mouthful.

### DIFF
--- a/Cargo.nix
+++ b/Cargo.nix
@@ -11,7 +11,7 @@
   lib,
 }:
 let
-  inherit (rustLib) fetchCratesIo fetchCrateLocal fetchCrateGit expandFeatures decideProfile genDrvsByProfile;
+  inherit (rustLib) fetchCratesIo fetchCrateLocal fetchCrateGit fetchCrateAlternativeRegistry expandFeatures decideProfile genDrvsByProfile;
   profilesByName = {
   };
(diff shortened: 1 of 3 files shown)


trha:
> we are solidifying that a single default.nix is sufficient for representing all crates within a single workspace

Yeah, because they share the same `Cargo.lock`.

> How would it look like to add multiple crates to the `default.nix`?
Hmm, I'm not sure what you mean. This should be the same as before, as this PR only provides a more convenient way to access workspace crates.

ebkalderon:
Thanks for discussing the pull request a bit more in-depth with me! I think it makes perfect sense to use a single `Cargo.nix` and top-level `default.nix` per workspace to mirror the single `Cargo.lock` file that they all share.

Just to make it clear to future readers: `cargo2nix` should be already be capable of transparently handling multi-crate workspaces identically to how it handles individual crate projects since it is only concerned with the `Cargo.lock` and top-level `Cargo.toml` manifest for certain details. This pull request just makes it easier to access all the crates inside a workspace as a single attribute set without having them be mixed in with the rest of the crate dependency graph, traversable via `rustPkgs.<registry>.<crate>."x.y.z"`.

workflow:
This is awesome @trha :purple_heart:

PR #47, opened 2019-12-10
Add runTests

trha:
This function runs all tests for a crate inside Nix.

### DIFF
--- a/overlay/run-tests.nix
+++ b/overlay/run-tests.nix
@@ -1,0 +1,19 @@
+# Run all tests for a crate.
+{ runCommand }:
+crate: env:
+let
+  testBins = crate { compileMode = "test"; };
+in
+  runCommand
+    "test-${testBins.name}"
+    ({ CARGO_MANIFEST_DIR = testBins.src; } // env)
+    ''
+        for f in ${testBins}/bin/*; do
+          # HACK: cargo produces the crate's main binary in the bin directory if the crate contains example tests.
+          # The `grep` filters out the main binary, which doesn't contain the help string found in test binaries.
+          if [[ -x "$f" ]] && grep "By default, all tests are run in parallel" "$f"; then
+            $f
+          fi
+        done
+        touch $out
+    ''
--- a/overlay/run-tests.nix
+++ b/overlay/run-tests.nix
@@ -6,7 +6,10 @@ let
 in
   runCommand
     "test-${testBins.name}"
-    ({ CARGO_MANIFEST_DIR = testBins.src; } // env)
+    ({
+      inherit (testBins) src;
+      CARGO_MANIFEST_DIR = testBins.src;
+    } // env)
     ''
         for f in ${testBins}/bin/*; do
           # HACK: cargo produces the crate's main binary in the bin directory if the crate contains example tests.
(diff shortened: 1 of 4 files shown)


ebkalderon:
To be clear, though, I feel like this would be useful for many simpler crates and as long as the hook isn't run by default, I believe it would be a fine feature to add to the `cargo2nix` overlay.

overlay/run-tests.nix
```diff
@@ -0,0 +1,22 @@
...
+    ''
+        for f in ${testBins}/bin/*; do
+          # HACK: cargo produces the crate's main binary in the bin directory if the crate contains example tests.
+          # The `grep` filters out the main binary, which doesn't contain the help string found in test binaries.
+          if [[ -x "$f" ]] && grep "By default, all tests are run in parallel" "$f"; then
+            $f
+          fi
+        done
+        touch $out
+    ''
```
snawaz: I assume.. in this case the tests should read messages (if any) using path `$CARGO_MANIFEST_DIR/messages/test_data.json` instead of `messages/test_data.json` because the latter wont work?

Also, it'd be good if you could use this: `RUST_BACKTRACE=1  $f  --nocapture`. It'd be also good if the function could accept a boolean to choose either `$f` (parallel) or `$f --test-threads=1`. Or maybe you accept number of threads. In `notifications`, I used single thread because my tests open files and there are lots of tests, as a result of which I get `too many open files`.

trha: Plain `messages/test_data.json` should work as `src` is specified.

> Also, it'd be good if you could use this: RUST_BACKTRACE=1 $f --nocapture

Good idea. I'll add a parameter to customize the test command.

snawaz: BTW, currently, I'm just creating a softlink before running the tests:
 ```bash
 ln -s $CARGO_MANIFEST_DIR/messages messages
 # for loop and run the tests
```

snawaz: > Plain `messages/test_data.json` should work as src is specified.

It does not. And I dont see why that should work. The tests dont care about `$CARGO_MANIFEST_DIR` unless you use that explicitly.

trha: This line
>        inherit (testBins) src; 
instructs `stdenv` to copy `src` to current working directory in the unpack phase.

trha: Added the ability to specify custom test command in https://github.com/tenx-tech/cargo2nix/pull/47/commits/4390238935c2fb527fc003342f6df06d451879be.

snawaz: Awesome. I was missing that line!

plumenator:
I sense an acute lack of documentation. Hopefully, this tool won't end up as impenetrable as the rest of the nix ecosystem. I think it would be great if we can document any "patterns" or utilities we implement.

plumenator:
To clarify: I meant documentation to aid discoverability. One could argue that we "just" need to look under `overlays/`, but that's not what a new user would do.

overlay/run-tests.nix
snawaz: > This line
> 
> > ```
> >    inherit (testBins) src; 
> > ```
> 
> instructs `stdenv` to copy `src` to the current working directory in the unpack phase.

If I understand it correctly, then `CARGO_MANIFEST_DIR` is not needed as such, except that it mimics `cargo`  behavior in this regard?

trha: Yes.

workflow:
@plumenator I agree! Got some time to help document it? :upside_down_face:

PR #50, opened 2019-12-15
Add simple command line arguments and Concise nix output

snawaz:
Review commit by commit for better experience.

- Add simple command line arguments
- Concise nix out
  - Use default values for many parameters to reduce the verbosity of the nix output.
  - Simplify `builtins.concatLists [ [a] [b] [c] ]` to `[a b c]` and such.
  - Reduce the output by up to 40% less lines.
- Bump version
- Print version as comment to the nix output


snawaz:
CI fails though:

```Rust
error: trait objects without an explicit `dyn` are deprecated
    --> src/buf/buf.rs:1154:30
     |
1154 | fn _assert_trait_object(_b: &Buf) {}
     |                              ^^^ help: use `dyn`: `dyn Buf`
     |
note: lint level defined here
    --> src/lib.rs:71:9
     |
71   | #![deny(warnings, missing_docs, missing_debug_implementations)]
     |         ^^^^^^^^
     = note: #[deny(bare_trait_objects)] implied by #[deny(warnings)]

error: trait objects without an explicit `dyn` are deprecated
    --> src/buf/buf_mut.rs:1167:30
     |
1167 | fn _assert_trait_object(_b: &BufMut) {}
     |                              ^^^^^^ help: use `dyn`: `dyn BufMut`

error: aborting due to 2 previous errors

error: Could not compile `bytes`.
```

Any idea how to fix this? 

_I dont have permission to rerun the CI btw_.

src/main.rs
```diff
@@ -459,31 +525,38 @@ impl<'a> ResolvedPackage<'a> {
...
             ] {
-                writeln!(f, "{} = {{", attr)?;
-
-                for ((dep_id, _), rdep) in self
+                let deps: Vec<_> = self
                     .deps
                     .iter()
                     .filter(|((_, dep_kind), _)| dep_kind == kind)
-                {
+                    .collect();
```
trha: Not a request change, but you can avoid materializing the iterator by using `iter.peekable().peek().is_none()`.

src/main.rs
```diff
@@ -32,6 +33,59 @@ type PackageName<'a> = &'a str;
...
+        io::Write::flush(&mut io::stdout().lock()).expect("flush stdout buffer");
+        let mut line = String::default();
+        io::stdin()
+            .read_line(&mut line)
+            .expect("failed to read input");
+        if line.trim() != "yes" {
+            println!("aborted!");
+        }
+    }
+    generate_cargo_nix(fs::File::create(&file).expect(&format!(
```
trha: Style: all `expect`s are in lower case, not title case.

src/main.rs
```diff
@@ -32,6 +33,59 @@ type PackageName<'a> = &'a str;
...
+        }
+    }
+}
+
+fn print_version() {
+    println!("cargo2nix-{}\n", env!("CARGO_PKG_VERSION"));
+}
+fn print_help() {
+    print_version();
+    println!("$ cargo2nix                     # Print the help");
```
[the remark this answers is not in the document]
snawaz: Oops it should have been `--stdout` instead of `--stdin`. If you want that to be removed, then why do you think printing the output to `stdout` is a good idea without `--stdout`?

src/main.rs
snawaz: Nice. I was looking for something like that... the `.collect()` seemed forced. Thanks.

src/main.rs
snawaz: 👍

overlay/mkcrate.nix
```diff
@@ -12,12 +12,12 @@
...
   version,
-  registry,
+  registry ? "registry+https://github.com/rust-lang/crates.io-index",
   src,
-  features,
-  dependencies,
-  devDependencies,
-  buildDependencies,
+  features ? [],
+  dependencies ? {},
```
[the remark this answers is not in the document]
snawaz: 👍

overlay/mkcrate.nix
```diff
@@ -12,12 +12,12 @@
...
   name,
   version,
-  registry,
+  registry ? "registry+https://github.com/rust-lang/crates.io-index",
   src,
-  features,
-  dependencies,
-  devDependencies,
-  buildDependencies,
+  features ? [],
```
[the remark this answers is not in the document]
snawaz: 👍

src/main.rs
```diff
@@ -32,6 +33,59 @@ type PackageName<'a> = &'a str;
...
+            println!("aborted!");
+        }
+    }
+    generate_cargo_nix(fs::File::create(&file).expect(&format!(
+        "Couldn't open file for writing: {}",
+        file.as_ref().display()
+    )));
+}
+
+fn generate_cargo_nix(mut out: impl io::Write) {
```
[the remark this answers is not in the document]
snawaz: I dont see that as necessary or need.. some function has to format `Display` right?

snawaz:
> > CI fails though:
> 
> This [change](https://github.com/tenx-tech/cargo2nix/pull/50/commits/f318c127b6af50c4e7e461e859d1ed6261c479e6#diff-05e469c68f7b6f9882bea7e5775c5602R15) defaulting `registry` to crates.io breaks the [`capLints` override](https://github.com/tenx-tech/cargo2nix/blob/master/overlay/overrides.nix#L73). Let's just remove it.

I've removed it. Hope it works. 

However, I didn't understand why it breaks `capLints`? Can we use default and still make `calLints` work? (Given that most projects use `crates.io` extensively and occasionally other registry).

src/main.rs
```diff
@@ -32,6 +33,59 @@ type PackageName<'a> = &'a str;
 type RootFeature<'a> = (PackageName<'a>, Feature<'a>);
 
 fn main() {
+    let args: Vec<String> = std::env::args().collect();
+    let args: Vec<&str> = args.iter().map(AsRef::as_ref).collect();
+
+    match args.as_slice() {
+        [_, "--stdin"] => generate_cargo_nix(io::stdout()),
```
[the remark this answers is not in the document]
snawaz: Added short names as well in the dumb way, and didn't use `structopt` (it looks complicated to me, so will refactor that part later).

src/main.rs
```diff
@@ -459,31 +525,38 @@ impl<'a> ResolvedPackage<'a> {
...
+                write!(f, "features = [")?;
+                for (feature, optionality) in self.features.iter() {
+                    let mut f = f.indent(2);
+                    match optionality
+                        .to_expr(outer.root_features, n_root_pkgs)
+                        .simplify()
+                    {
+                        True => write!(f, " {:?}", feature)?,
+                        expr => {
+                            write!(f, " ({} ({}) {:?})", outer.optional, expr.to_nix(), feature)?
```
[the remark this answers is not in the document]
snawaz: Can we avoid printing the `null` to begin with?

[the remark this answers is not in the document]
snawaz: Used `concatLists` .. until we avoid printing `null`.

src/main.rs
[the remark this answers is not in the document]
snawaz: Changed `--stdin` to `--stdout` only... and resolving this issue until you disagree and wanna discuss it.

snawaz:
@trha .. addressed the feedbacks: see the latest commit.

src/main.rs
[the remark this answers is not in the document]
trha: At least move the `Generated by ...` part inside the `display` block above.

src/main.rs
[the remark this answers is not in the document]
trha: We can't avoid `null`s. Features of dependencies depend on root features, which are specified at initialization time.
Also, please return the old formatting.

src/main.rs
```diff
@@ -32,6 +33,64 @@ type PackageName<'a> = &'a str;
...
+            "warning: do you want to overwrite '{}'? yes/no: ",
+            file.as_ref().display()
+        );
+        io::Write::flush(&mut io::stdout().lock()).expect("flush stdout buffer");
+        let mut line = String::default();
+        io::stdin()
+            .read_line(&mut line)
+            .expect("failed to read input");
+        if line.trim() != "yes" {
+            println!("aborted!");
```
trha: Required.
There should be a `return` here.

src/main.rs
```diff
@@ -32,6 +33,64 @@ type PackageName<'a> = &'a str;
...
+    println!("$ cargo2nix                        # Print the help");
+    println!("$ cargo2nix -s,--stdout            # Output to stdout");
+    println!("$ cargo2nix -f,--file              # Output to Cargo.nix");
+    println!("$ cargo2nix -f,--file <file>       # Output to the given file");
+    println!("$ cargo2nix -v,--version           # Print version of cargo2nix");
+    println!("$ cargo2nix -h,--help              # Print the help");
+}
+
+fn write_to_file(file: impl AsRef<Path>) {
+    if file.as_ref().exists() {
```
trha: Let's avoid the repeated `as_ref`:
```suggestion
    let file = file.as_ref();
    if file.exists() {
```

src/main.rs
```diff
@@ -32,6 +33,64 @@ type PackageName<'a> = &'a str;
...
+        print!(
+            "warning: do you want to overwrite '{}'? yes/no: ",
+            file.as_ref().display()
+        );
+        io::Write::flush(&mut io::stdout().lock()).expect("flush stdout buffer");
```
trha: This lock is unnecessary.
```suggestion
        io::Write::flush(&mut io::stdout()).expect("flush stdout buffer");
```

src/main.rs
```diff
@@ -32,6 +33,64 @@ type PackageName<'a> = &'a str;
...
+        let mut line = String::default();
```
trha:
```suggestion
        let mut line = String::new();
```

src/main.rs
```diff
@@ -32,6 +33,64 @@ type PackageName<'a> = &'a str;
 type RootFeature<'a> = (PackageName<'a>, Feature<'a>);
 
 fn main() {
+    let args: Vec<String> = std::env::args().collect();
+    let args: Vec<&str> = args.iter().map(AsRef::as_ref).collect();
+
+    match args.as_slice() {
+        [_, "--stdout"] => generate_cargo_nix(io::stdout()),
```
trha: Use `lock` here for better performance.
```suggestion
        [_, "--stdout"] => generate_cargo_nix(io::stdout().lock()),
```

trha: You can also merge similar branches
```suggestion
        [_, "--stdout"] | [_, "-s"] => generate_cargo_nix(io::stdout()),
```

src/main.rs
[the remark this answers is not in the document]
snawaz: Moved that inside.

Also, added `cargo2nixVersion = <version>;`. I think we should use this variable to ensure that we use `import cargo2nix` package which is compatible with `Cargo.nix`.

src/main.rs
snawaz: Yes. Thanks.

src/main.rs
```diff
@@ -459,31 +527,38 @@ impl<'a> ResolvedPackage<'a> {
...
-            writeln!(f, "features = builtins.concatLists [")?;
-            for (feature, optionality) in self.features.iter() {
-                let mut f = f.indent(2);
-                match optionality
-                    .to_expr(outer.root_features, n_root_pkgs)
-                    .simplify()
-                {
-                    True => writeln!(f, "[ {:?} ]", feature)?,
-                    expr => writeln!(f, "({} ({}) {:?})", outer.optional, expr.to_nix(), feature)?,
+            if self.features.len() != 0 {
```
trha:
```suggestion
            if !self.features.is_empty() {
```

src/main.rs
snawaz: Done.

src/main.rs
snawaz: Done

src/main.rs
snawaz: Done

trha:
> Can we use default and still make `capLints` work?

We can, but I can't think of any way that doesn't add complexity to the overlay, so let's just leave it there for now.

trha:
You might need to update `mkcrate-nobuild.nix` as well. Does `nix-shell` work on this branch?

src/main.rs
snawaz: That's nice. I was expecting something like that to exist.. and I tried something, and it didn't work.. and I gave up.

Thanks.

src/main.rs
snawaz: ah. I tried `empty()`.. did not it is `is_*` form.

snawaz:
> You might need to update `mkcrate-nobuild.nix` as well. Does `nix-shell` work on this branch?

No. It didn't. Fixed that too. Thanks.

snawaz:
> Changes not marked "Required" are optional. Can you also split out the `localPatterns` fix to another PR and restore the old formatting for `features`?

Is there any visible cue to _Required_ changes? I dont seem to see any.. 😕 

- I've removed the `localPatterns` fix from this PR.
- I think the current formatting for `features` is good, given that most of the times it has only fewer items in it. Can we fix that if we see that as problem in future?

snawaz:
> One last change: Please bring back the old formatting for `features`, or lines like [this](https://github.com/tenx-tech/quotes/blob/cargo2nix/Cargo.nix#L94) would be too long.

ah. I see. Changed to the old formatting. (We could still make it better, but it is not worth doing now).

src/main.rs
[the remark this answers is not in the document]
snawaz: Done.

src/main.rs
```diff
@@ -459,31 +524,38 @@ impl<'a> ResolvedPackage<'a> {
...
-                    expr => writeln!(f, "({} ({}) {:?})", outer.optional, expr.to_nix(), feature)?,
+            if !self.features.is_empty() {
+                writeln!(f, "features = builtins.concatLists [")?;
+                for (feature, optionality) in self.features.iter() {
+                    let mut f = f.indent(2);
+                    match optionality
+                        .to_expr(outer.root_features, n_root_pkgs)
+                        .simplify()
+                    {
+                        True => writeln!(f, "[{:?}]", feature)?,
```
trha:
```suggestion
                        True => writeln!(f, "[ {:?} ]", feature)?,
```

snawaz: Fixed.

snawaz:
> Thanks for the improvements! 1170 lines fewer is huge! The next step would be adding a warning to `makePackageSet` if the overlay version and the code generator version differs.

Yes. Have created the issue for that.

PR #56, opened 2020-01-08
Exclude generated Cargo.nix from the repo language stat

eupn:
This will make language stat of the repo more sensible.


eupn:
Github doesn't seem to catch up with the change yet, the language statistic is the same. Probably it will update later when the repo is re-indexed again by the Github.

eupn:
Stats are updated!

PR #57, opened 2020-01-10
Less verbose cargo

eupn:
During the work on Cumbertrade I've noticed that `cargo` is producing a lot of noise during a build of `cargo2nix`-ed packages. Related issue is here: https://github.com/tenx-tech/engineering-program/issues/721

This PR won't just remove the `-vvv` flag from `cargo` invocation in `mkRustCrate`, but will make the verbosity level configurable since I believe that this high verbosity was added for a reason and still might be useful in some cases.

cc @workflow @ebkalderon

### DIFF
--- a/overlay/mkcrate.nix
+++ b/overlay/mkcrate.nix
@@ -23,6 +23,7 @@
   meta ? { },
   rustcflags ? [ ],
   rustcBuildFlags ? [ ],
+  NIX_DEBUG ? 0,
 }:
 with builtins; with lib;
 let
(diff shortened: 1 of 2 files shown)


trha:
Let's respect nixpkgs [`NIX_DEBUG` variable](https://nixos.org/nixpkgs/manual/#ssec-stdenv-attributes). User can enable verbose build per crate using overrides.

ebkalderon:
I like the idea behind this PR! Let's respect `NIX_DEBUG` as well, as @trha suggested.

eupn:
@trha @ebkalderon I've tried this approach, please take a look!

trha:
We actually don't need to change code generation for this. User can specify `NIX_DEBUG` for all crates using:
```
  packageOverrides = pkgs: [
    (pkgs.rustBuilder.rustLib.makeOverride { 
      overrideAttrs = _: { NIX_DEBUG = "7"; }; 
    })
  ];
};
```

eupn:
@trha please check

PR #64, opened 2020-01-21
Fix building lib targets with non-standard names

ebkalderon:
### Fixed

* Fix building `[lib]` targets with non-standard names.

Currently, `cargo2nix` assumes that all crate library targets follow the format of `lib<crate_name_with_underscores>.rlib`, which isn't always true. It turns out that the library target of the [`sha-1`] crate uses a non-standard rlib name which doesn't match the name of the crate, thanks to the `lib.name` attribute in the `Cargo.toml`. Because of this, the `sha-1` crate doesn't build correctly in `cargo2nix`.

[`sha-1`]: https://github.com/RustCrypto/hashes/tree/master/sha1

This commit fixes the build by first checking whether the `lib.name` attribute exists in the crate's `Cargo.toml` first, and if it doesn't exist, it will fall back to `crateName`.

CC @port3000

### DIFF
--- a/overlay/mkcrate.nix
+++ b/overlay/mkcrate.nix
@@ -95,7 +95,6 @@ let
     inherit NIX_DEBUG;
     name = "crate-${name}-${version}${optionalString (compileMode != "build") "-${compileMode}"}";
     inherit src version meta;
-    crateName = replaceChars ["-"] ["_"] name;
     buildInputs = runtimeDependencies;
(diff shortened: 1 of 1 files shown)


ebkalderon:
@trha I think you meant #61, haha. 😉 I feel I should've been more careful when reviewing the conversion from Nix to shell script with `remarshall` in that pull request, but it's okay. I'm glad it was isolated quickly in the end!

PR #67, opened 2020-02-06
Upgrade to cargo 0.41.0 and release cargo2nix 0.8.0

ebkalderon:
### Added

* Add dependency on `cargo_platform` 0.1.0.

### Changed

* Upgrade `cargo` to version 0.41.0.
* Bump `cargo2nix` crate version to 0.8.0.

These changes are necessary to get the `cargo2nix -f` command to successfully parse the newest `Cargo.lock` file from [rust-analyzer].

[rust-analyzer]: https://github.com/rust-analyzer/rust-analyzer/


ebkalderon:
@trha Yeah, there was a new diff-friendly `Cargo.lock` format introduced recently, and it seems some prominent projects have already switched over to it.

I could have migrated `cargo2nix` to use `cargo` 0.42.0 (the latest version) as well, but I instead chose version 0.41.0 because that is the last version that can still be built with Rust 1.34. Unfortunately, `cargo` 0.42.0 requires at least Rust 1.40 because of its use of `std::mem::take()`.

Issue #51, opened 2019-12-16
Check version compatibility between cargo2nix and Cargo.nix


snawaz:
It'd be good to ensure that the imported `cargo2nix` is compatible with `Cargo.nix` of the project.

- `Cargo.nix` exposes [`cargo2nixVersion`][1] (the version of `cargo2nix` which was used to generate this file). We could use this to ensure that the imported `cargo2nix` version is compatible, else it should fail with a good error message.

[1]: https://github.com/tenx-tech/cargo2nix/pull/50#discussion_r358304625

ebkalderon:
Just to clarify, when you say "imported `cargo2nix`," I assume you mean the imported `cargo2nix` overlay?

I think this is definitely doable! It would mean that `makePackageSet` would need to check the version specified by `cargo2nixVersion` against its own and `throw` an error if they are incompatible. With that said, I would much prefer it if we didn't use a straight string equality check for this task and instead checked for semantic versioning compatibility between them. Perhaps something like [`lib.versionAtLeast`](https://github.com/NixOS/nixpkgs/blob/master/lib/strings.nix#L473) or [`builtins.compareVersions`](https://nixos.org/nix/manual/#ssec-builtins) could help here.

snawaz:
> Just to clarify, when you say "imported cargo2nix," I assume you mean the imported cargo2nix overlay?

Yes, the overlays &mdash; the nix stuff the clients could use.

psionic-k:
I could be wrong, but I don't think there's anything left to do on this issue

PR #78, opened 2020-02-16
Enforce version requirement for cargo2nix to avoid accidental downgrade

snawaz:
Multiple developers working on a project may lead to a situation in which one developer accidentally downgrades `Cargo.nix` due to regeneration by an older `cargo2nix` available on his/her system. This PR is all about avoiding that. It detects such a case and then exits with the following error message:

image

Hope the idea of color coding isn't bad.

- Incremented `Cargo.toml` version.
- Regenerated `Cargo.nix` using _this_ version.
- Little bit of refactoring.

Also, we don't need to support downgrade because in case if users want to downgrade, they can delete the existing `Cargo.nix` and regenerate using an older version. The above error generation is possible only if this file exists!


ebkalderon:
Thank you very much for the pull request, @snawaz! The color coding for the text is fine, I think, although depending on how it is implemented, it might potentially mess with stdio rendering on certain CI platforms.

I am not sure we should be throwing an error on patch version incompatibilities for `cargo2nix`, as this is a pretty big pessimization. We don't have an officially documented semantic versioning policy (we should definitely add one), but at present:

1. Any change made to the expression structure in the `Cargo.nix` would be at least require a minor version bump, since we are currently in the `0.x.y` series. These changes should likely be denied by the logic in this PR.
2. Patch versions indicate some minor backwards-compatible addition or bugfix, e.g. the addition of a new crate override to the overlay. Changes like these do not affect `Cargo.nix` compatibility, and I don't believe it makes sense to deny users from regenerating the file based on a patch-based version difference.

I think this would provide users with reasonable compatibility guarantees until we reach 1.0, at which point we should probably consider stabilizing the `Cargo.nix` expression format prior to release.

This pull request is a great start, but I think we should be a bit more lenient about patch version differences since we don't want to over-aggressively interrupt user workflows over ultimately inconsequential changes.

snawaz:
> 1. Any change made to the expression structure in the `Cargo.nix` would be at least require a minor version bump, since we are currently in the `0.x.y` series. These changes should likely be denied by the logic in this PR.

That's a good/desirable thing then? _assuming I understood your point correctly_.

> Patch versions indicate some minor backwards-compatible addition or bugfix, e.g. the addition of a new crate override to the overlay. Changes like these do not affect Cargo.nix compatibility, and I don't believe it makes sense to deny users from regenerating the file based on a patch-based version difference.

_bugfix_. That's it. It is always good to be aware/explicit about the transition from _buxfixed state_ to _bug state_.  I mean, the transition from `x.y.4` to `x.y.5` should be implicit, while the reverse should be explicit (because that compromises the bugfixes).

Also, as said earlier, the users can regenerate the file if they want to downgrade. The only extra step is `rm -f Cargo.nix` before the regeneration. But I believe most often users dont want to downgrade, hence the error is the default. It's all about _being explict_.

ebkalderon:
In regards to the first point, yes, that's indeed the behavior we want. We should definitely retain that aspect of this PR!

But one should keep in mind that bugfixes that result in a minor patch version do not result in any change in the `Cargo.nix` format (see the first point in the list). If a bug is fixed regarding the `cargo2nix` CLI tool and results in a breaking change in the `Cargo.nix` format, then it will be released as a minor version bump.

I suggest that bugfixes made to the `cargo2nix` overlay or to the binary that do not affect the Nix expression output of `cargo2nix -f` (patch version changes) do not need to be denied. Any bugfixes which alter the output of the `Cargo.nix` in a breaking way such that evaluation fails or our feature or platform resolution logic changes (minor version changes) should be denied.

snawaz:
@ebkalderon 

What you said makes sense ideally. However, my concern was if some incompatible format/requirements slips though the _patch_ version. Maybe we should add a _checklist_ for each commit, like we have for our internal project mothership?

Anyway, I've relaxed the requirement. Have a look at the recent commit (`Relax the version requirement to major and minor, ignoring the patch`). 

Also, see the UI for both _match success_ and _match failure_.

image

ebkalderon:
@snawaz I think we can probably do that for the short term, but it would be very nice if CI could catch this as well in the future.

One thing that @trha suggested to me is whether or not we could opt to have the Nix overlay check for compatibility at build time instead of at codegen time (#51). The implementation is simpler, since Nix itself provides a `builtins.compareVersions` function for checking compatibility between two version strings, and CI will deny any pull requests containing breaking changes to the `Cargo.nix` file. There is also no need to scan through the `Cargo.nix` source line-by-line to identify the value of the `cargo2nixVersion` string. What do you think about this?

snawaz:
I think earlier we check, better we'll be. Like static typing vs dynamic typing. So checking at codegen is like static checking. The CI could do check as well &mdash; especially the things we cannot do at codegen time.

Also, _our_ CI is internal thing, specific to TenX? How do we enforce it outside?

_BTW, I just fixed the issue finally (rebased it). The earlier commit had problem_

ebkalderon:
When I say that CI could block incompatible builds, I mean that any attempt to call `nix-build` on any `cargo2nix` project, first or third party, with an incompatible `Cargo.nix` file will throw an evaluation time error. Any CI pipeline which attempts to build the project itself will fail immediately before the build starts.

I guess I don't mind performing both kinds of checks for maximum strictness, though the static check adds much more overhead to the Rust code than the equivalent check in the Nix overlay code. The only difference is that the codegen check catches it when running `cargo2nix -f`, whereas the overlay check catches it when running `nix-build`.

snawaz:
> When I say that CI could block incompatible builds, I mean that any attempt to call nix-build on any cargo2nix project..

I see. Yes, we want to do this as well. I had a small discussion with @trha regarding this when I had added the version attribute to the codegen. 

I think we should have both. They're not mutually exclusive. Catching some problems at codegen-time helps understand the problem _quickly_ &mdash; else we have to come through all the way from _nix-build_ to _parsing nix errors_ to _examining and understanding them_ to _codegen step_ eventually, only to realize that we used incompatible `cargo2nix` version to regenerate.

ebkalderon:
Right! I suppose we can have both implemented. Let's see if we can solve #51 in another PR separately, and maybe further streamline this code at a later date. 😄

PR #79, opened 2020-02-17
Set PKG_CONFIG_ALLOW_CROSS if host and build platforms differ

ebkalderon:
The [`pkg-config`](https://crates.io/crates/pkg-config) crate cannot cross-compile packages unless the `PKG_CONFIG_ALLOW_CROSS` environment variable is set, throwing the [`CrossCompilation`] error otherwise.

[`CrossCompilation`]: https://docs.rs/pkg-config/0.3.17/pkg_config/enum.Error.html#variant.CrossCompilation

This commit automatically sets `PKG_CONFIG_ALLOW_CROSS=1` as a propagated environment variable for any dependency tree which depends on `pkg-config` if the `stdenv` host platform and build platform differ.

This should fix cross-compilation of certain crates which rely on `pkg-config`, such as [`curl-sys`](https://crates.io/crates/curl-sys).

### DIFF
--- a/overlay/overrides.nix
+++ b/overlay/overrides.nix
@@ -64,6 +64,7 @@ in rec {
     fsevent-sys
     libgit2-sys
     openssl-sys
+    pkg-config
     pq-sys
     prost-build
     rand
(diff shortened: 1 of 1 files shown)


ebkalderon:
Currently blocking on #78, since that's a much bigger and more important PR and this should be easy to rebase.

ebkalderon:
Okay, rebased! This PR should be ready to review, @trha.

overlay/overrides.nix
```diff
@@ -118,6 +119,19 @@ in rec {
     };
   };
 
+  pkg-config = if pkgs.stdenv.hostPlatform == pkgs.stdenv.buildPlatform
```
trha: Shouldn't this be
```suggestion
  pkg-config = if pkgs.stdenv.hostPlatform != pkgs.stdenv.buildPlatform
```

ebkalderon: D'oh! :man_facepalming: That is precisely what it should be. Good catch there. Thank you!

PR #83, opened 2020-02-25
Replace is_valid_ident() regex with iterator

ebkalderon:
### Changed

* Replace `is_valid_ident()` regular expression with some simple iterator code .

### Removed

* Remove dependencies on `regex` and `once_cell`.

Pulling in `regex`, a full-featured regular expression engine, is a bit overkill when this could just be represented in a few lines of `chars()` iterators. Hopefully this saves a bit of compile-time and run-time performance as well.


ebkalderon:
Sorry about that, @eupn! I had forgotten to actually remove the `regex` dependency from the `Cargo.toml` and regenerate the `Cargo.nix`. Kindly re-requesting your review!

ebkalderon:
I've rebased against the latest `master` after merging #84.

src/expr.rs
```diff
@@ -163,10 +160,14 @@ where
...
 
 fn is_valid_ident(id: &str) -> bool {
-    static IDENT_RE: OnceCell<Regex> = OnceCell::new();
-    IDENT_RE
-        .get_or_init(|| Regex::new(r"^[a-zA-Z_][a-zA-Z0-9_'-.]*$").unwrap())
-        .is_match(id)
+    let mut chars = id.chars();
+
+    match chars.next() {
+        Some(c) if c.is_ascii_alphanumeric() || c == '_' => {}
```
snawaz: You intended to use `is_ascii_alphabetic` instead of `is_ascii_alphanumeric`? Because the latter would make `2abc` a valid identifier? I'm sure you didn't mean that.

snawaz: BTW, dow does this look: 

```rust
id.chars()
  .nth(0)
  .map(|c| c.is_ascii_alphanumeric() || c == '_' )
  .unwrap_or(false) && chars.all(|c| c.is_ascii_alphanumeric() || "_'-.".contains(c))
```
? 🤔 

Does it need to be `mut`?

ebkalderon: Yes, good catch! I did intend to use `is_ascii_alphabetic()` in this particular case. I'll revise the commit and fix it.

That one-liner is almost correct, you would only have to add `.skip(1)` before calling `all()` in the second branch to ensure you do not evaluate the first character.

Yes, `chars` does need to be `mut` if you wish to call `nth()` or `next()` on it because those methods take `&mut self`.

ebkalderon: I'm of a mixed opinion as to how much more readable the combinator-heavy one-liner makes the code compared to just calling `next()` on the first element. :confused:

ebkalderon: I'm of a mixed opinion as to how much more readable the combinator-heavy one-liner makes the code compared to just calling `next()` on the first element. :confused:

snawaz: > If we end up using `id.chars()` again in the second branch instead of reusing `chars`, we would need to call `skip(1)` to ensure that `all()` doesn't include the first character.

You dont have to use `skip(1)` actually because the first character is a subset of the remaining characters... so it doesn't matter if you process the first character twice!

ebkalderon: That's correct, but that's technically not what the original regex was doing. And you're evaluating a character twice needlessly. 😄

snawaz: Haha. It simplifies things without much cost. :D

snawaz: Anyway, I consider such things a matter of taste. So I'm okay with either way. As long as it works, and relatively easy to read, everything is fine. Haha.

PR #84, opened 2020-02-25
Ensure Cargo.nix file is written atomically

### LINKED ISSUE #76: Empty Cargo.nix file is generated upon panic
It appears that when `cargo2nix -f` is run against a crate workspace and Cargo aborts execution due to `--locked` (see #75), a zero-length `Cargo.nix` file is inadvertently produced. It seems that we call `File::create()` too early in the startup phase, creating an empty file that ends up being left behind after `cargo2nix` exits with an error.

ebkalderon:
### Fixed

* Fix bugs associated with non-atomic writing of the `Cargo.nix` file.

With this commit, we now write the `Cargo.nix` file to a temporary file first and atomically `mv` it to the correct location once the templating successful. This should ensure that if the `cargo2nix` command panics or otherwise exits prematurely with an error, the existing `Cargo.nix` file for the project is not accidentally truncated.

This introduces a new dependency on `tempfile`, but this could be removed again with a bit of strategic refactoring later.

Closes #76.


src/main.rs
```diff
@@ -144,10 +144,16 @@ fn write_to_file(file: impl AsRef<Path>) {
...
+
+    let mut temp_file = tempfile::Builder::new()
+        .tempfile()
+        .expect("could not create new temporary file");
+
+    generate_cargo_nix(&mut temp_file);
+
+    temp_file
+        .persist(path)
+        .unwrap_or_else(|_| panic!("could not write file to final location: {}", path.display()));
```
snawaz: Regarding the `_` thingy: maybe we should print the error instead of ignoring it?

snawaz: Regarding the `_` thingy: maybe we should print the error instead of ignoring it?

ebkalderon:
@snawaz The reason I went with `tempfile::Builder` is because it returns a `tempfile::NamedTempFile`, which can be persisted to a permanent path, while `tempfile::tempfile()` returns a regular `std::fs::File` which cannot be persisted to a path.

src/main.rs
ebkalderon: I decided not to print the error here since we had ignored it in the original code with `expect()` and I wanted to do the same. If you'd like for me to print the error, I'd be happy to add another commit!

BTW, I chose to use `unwrap_or_else()` here instead of `expect()` to avoid calling `format!()` needlessly in the happy path. This is essentially equivalent to `expect("<message>")`.

snawaz: I think printing the error could be helpful if it fails at all. There doesn't seem to be any harm, at the least. :-)

And yes, I understood `unwrap_or_else()` vs `expect()` .. it is minor (and personal preference), though I personally prefer to use the latter because it is shorter (and performance is not required in such cases, or have negligible impact).

ebkalderon: No worries, I think I'll add a commit printing the error in this case. 😄

ebkalderon: Done! 👍

PR #85, opened 2020-02-26
Clean up some workspace resolution code

ebkalderon:
### Changed

 * Simplify `resolve_ws_with_opts()` invocations in several methods by using `ResolveOpts::everything()` and `ResolveOpts::new()` to reduce boilerplate.
* Remove explicit return in `all_eq()`.
* Remove explicit lifetimes from `all_features()`.
* Add single trailing newline for readability.
* Keep data structures close to their `impl` blocks.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -6,14 +6,13 @@ use std::{
     fs,
     io::{self, BufRead},
     path::Path,
-    rc::Rc,
 };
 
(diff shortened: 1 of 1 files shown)


ebkalderon:
I think we should expect a new GitHub release after this, and then some more deep refactoring, documentation, and cleanups.

PR #86, opened 2020-02-26
Improve Rust error handling

### LINKED ISSUE #81: Improve Rust error handling
The original code for the `cargo2nix` CLI tool was a quick prototype and therefore does not properly handle errors with `Result`. We should refactor the `main.rs` and all associated modules to use proper error handling instead of panicking.

ebkalderon:
### Changed

* Use `failure` as application-wide error type, matching the `cargo` crate which we link to.
* Restructure code to remove all panics, unwraps, and expects _except_ those which are guaranteed to not occur in practice (meaning that if they occur, something is seriously wrong).
* Print all error messages to `stderr` instead of `stdout`.
* Compress extra newlines in message output a bit to match the visual style of upstream Cargo.
* Minor cosmetic cleanups, e.g. removal of some unnecessary turbofishes.

Sorry for the very large pull request, everyone! I'd recommend going commit-by-commit for this one, and please feel free to ask questions if you have them. This isn't intended to be perfect; there will be more refinements to come.

Closes #81.


src/main.rs
```diff
@@ -33,98 +34,110 @@ mod template;
...
 const VERSION_ATTRIBUTE_NAME: &str = "cargo2nixVersion";
 
 fn main() {
     let args: Vec<String> = std::env::args().collect();
     let args: Vec<&str> = args.iter().map(AsRef::as_ref).collect();
+    if let Err(err) = try_main(&args) {
+        eprint!(colorify!(red_bold: "error: "));
+        eprintln!("{}", err);
+        std::process::exit(1);
+    }
```
snawaz: non-issue: just wondering if we can simply write `fn main() -> Result<(), Error>`?

src/main.rs
```diff
@@ -33,98 +34,110 @@ mod template;
...
+            print_help()
         }
     }
 }
 
 fn version() -> Version {
-    Version::parse(env!("CARGO_PKG_VERSION")).expect("parse CARGO_PKG_VERSION")
+    // Since `CARGO_PKG_VERSION` is provided by Cargo itself, which uses the same `semver` crate to
+    // parse version strings, the `unwrap()` below should never fail.
+    Version::parse(env!("CARGO_PKG_VERSION")).unwrap()
```
snawaz: I think `unwrap()` and `expect()` both are fundamentally the same thing, with only difference is that the latter takes a user-provided string (which is a good documentation as well). 

Also, both are a form of _assert_; so if your assertion is that `unwrap()` should never fail, then _that_ assertion doesn't change if you use `expect()` instead (I mean you can still make the same assertion). The latter, however, has one added advantage: IF your assertion fails, you'll get a better error message. :-)

If that makes sense, I'd suggest to revert this change. :-)

src/main.rs
```diff
@@ -33,98 +34,110 @@ mod template;
...
 }
 
-fn read_version_attribute(path: &Path) -> Version {
-    let file = fs::File::open(path).expect(&format!("Couldn't open file {}", path.display()));
+fn read_version_attribute(path: &Path) -> Result<Version> {
+    let file = fs::File::open(path).context(format!("Couldn't open file {}", path.display()))?;
```
snawaz: We're ignoring the cause of the failure here. So maybe, we should use `with_context(|e| ..)`?

src/main.rs
```diff
@@ -33,98 +34,110 @@ mod template;
...
     io::BufReader::new(file)
         .lines()
         .filter_map(|line| line.ok())
         .find(|line| line.trim_start().starts_with(VERSION_ATTRIBUTE_NAME))
         .and_then(|s| {
             if let Some(i) = s.find('"') {
                 if let Some(j) = s.rfind('"') {
-                    return Some(Version::parse(&s[i + 1..j]).expect("parse version attribute"));
+                    return Version::parse(&s[i + 1..j]).ok();
```
snawaz: I think we should use `Result<>` instead of `Option<>` here so that we dont have to ignore the parsing error?

src/main.rs
```diff
@@ -33,98 +34,110 @@ mod template;
...
-            requirement,
-            path.display()
-        )),
-        ver,
-    )
+fn version_req(path: &Path) -> Result<(VersionReq, Version)> {
+    let version = read_version_attribute(path)?;
+    let requirement = format!(">={}.{}", version.major, version.minor);
+    VersionReq::parse(&requirement)
+        .map_err(|_| failure::format_err!("parse {} found in {}", requirement, path.display()))
```
snawaz: Print `|e|`?

src/main.rs
ebkalderon: As described in the first commit message (99871bc), the issue is that having `main()` return a `Result` is that it prints the `Debug` representation of the error instead of `Display`, which results in a poor experience for the end-user as a CLI tool.

src/main.rs
ebkalderon: That makes sense in general sure! 👍 But the only way that this would fail is if Cargo itself is broken, at which point, you would have bigger problems than a stock panic message. I feel that the addition of the message is redundant here for this reason.

src/main.rs
ebkalderon: We didn't print the cause in the original code either, but now that you mention it, I think we probably should for `File::open()`. I'll add it in a subsequent commit! 👍

src/main.rs
ebkalderon: I initially considered that when working on this commit, but the problem is that the final output type of the combinator chain is `Result<Result<_, _>, _>`, which results in a somewhat clunky appearance. 😞 In this case, however, the error isn't actually being ignored as a `None` will trigger the `Valid cargo2nixVersion not found in {}` error.

I do agree that it's not as informative, though. I think I'll refactor this block in a subsequent PR to reduce the combinator nesting and deep rightward drift. I'll re-introduce the correct error reporting then.

src/main.rs
ebkalderon: Sure, I think I can try to print the cause as well.

src/main.rs
trha: `context` already records the error cause in the newly returned error. The recommended way to print the whole error chain is to use [`iter_chain`](https://docs.rs/failure/0.1.6/failure/struct.Error.html#method.iter_chain), similar to what Cargo [does](https://docs.rs/cargo/0.42.0/src/cargo/util/errors.rs.html#403-409). In fact I think we should reuse that function from Cargo.

src/main.rs
trha: Can we use [`context`](https://docs.rs/failure/0.1.6/failure/trait.ResultExt.html#tymethod.context) here too?

src/main.rs
ebkalderon: @trha That's a great idea! I'll update the latest commit to do just that.

src/main.rs
ebkalderon: Yup, I'll revise the latest commit to do just that.

src/main.rs
```diff
@@ -33,98 +34,112 @@ mod template;
...
+    // Since `CARGO_PKG_VERSION` is provided by Cargo itself, which uses the same `semver` crate to
+    // parse version strings, the `unwrap()` below should never fail.
+    Version::parse(env!("CARGO_PKG_VERSION")).unwrap()
 }
 
-fn read_version_attribute(path: &Path) -> Version {
-    let file = fs::File::open(path).expect(&format!("Couldn't open file {}", path.display()));
+fn read_version_attribute(path: &Path) -> Result<Version> {
+    let file = fs::File::open(path)
+        .with_context(|e| format!("Couldn't open file {}: {}", path.display(), e))?;
```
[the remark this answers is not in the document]
ebkalderon: Sounds good. I read the earlier comments; I'll update that in the next commit!

src/main.rs
```diff
@@ -133,82 +148,82 @@ fn write_to_file(file: impl AsRef<Path>) {
...
-        .expect("could not create new temporary file");
+        .context("could not create new temporary file")?;
 
-    generate_cargo_nix(&mut temp_file);
+    generate_cargo_nix(&mut temp_file)?;
 
     temp_file
         .persist(path)
-        .unwrap_or_else(|e| panic!("could not write file to {}: {}", path.display(), e));
+        .with_context(|e| format!("could not write file to {}: {}", path.display(), e))?;
```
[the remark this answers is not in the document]
ebkalderon: Will do!

src/main.rs
```diff
@@ -33,98 +34,112 @@ mod template;
...
+type Result<T> = std::result::Result<T, Error>;
 
 const VERSION_ATTRIBUTE_NAME: &str = "cargo2nixVersion";
 
 fn main() {
     let args: Vec<String> = std::env::args().collect();
     let args: Vec<&str> = args.iter().map(AsRef::as_ref).collect();
+    if let Err(err) = try_main(&args) {
+        eprint!(colorify!(red_bold: "error: "));
+        eprintln!("{}", err);
```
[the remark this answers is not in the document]
ebkalderon: Yeah, I'll update that. I'm quite happy that `display_causes()` exists.

src/main.rs
```diff
@@ -390,41 +406,45 @@ impl<'a> ResolvedPackage<'a> {
...
+            let source_id = pkg.package_id().source_id();
+            if checksum.is_none() && source_id.is_git() {
+                let url = source_id.url().as_str();
+                let rev = source_id.precise().ok_or_else(|| {
+                    failure::format_err!("no precise git reference for {}", pkg.package_id())
+                })?;
+                prefetch_git(url, rev)
+                    .map(Cow::Owned)
+                    .map(Some)
+                    .with_context(|e| {
```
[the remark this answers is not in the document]
ebkalderon: Sure, I'll change that.

src/main.rs
```diff
@@ -33,98 +34,112 @@ mod template;
...
-            requirement,
-            path.display()
-        )),
-        ver,
-    )
+fn version_req(path: &Path) -> Result<(VersionReq, Version)> {
+    let version = read_version_attribute(path)?;
+    let req = format!(">={}.{}", version.major, version.minor);
+    VersionReq::parse(&req)
+        .map_err(|e| failure::format_err!("parse {} found in {}: {}", req, path.display(), e))
```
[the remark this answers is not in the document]
ebkalderon: Sure, I'll change that.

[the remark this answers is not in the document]
trha: IIRC `Error: From<Context<D>>` so we can use the `?` operator here to coerce the error type.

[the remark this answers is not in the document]
ebkalderon: Hmm, I'd like to avoid using `?` to keep the expression-like appearance of 104, but I just realized that we could do `map_err(Error::from)` after calling `context()`, so it's fine. 😄

ebkalderon:
Okay, I've switched it over to use `context()` and `display_causes()` where possible. It looks much better than having to manage the error causes by hand! Thanks for the suggestion.

ebkalderon:
No worries! I think I'll wait a bit before merging for @eupn and @snawaz to weigh in as well, in case they have any further suggestions of their own.

src/main.rs
snawaz: The meaning of *assertion* is that the programmer thinks something is impossible. `unwrap()` and `expect()` both convey that. So if and when that assumption turned out to be wrong, then of course you have a _bigger_ program, irrespective of whichever assertion form you use, but then you'll not have a custom message to quickly narrow down the issue if you use `unwrap()`. That's my opinion.

Also, if a programmer thinks something is not impossible, then he/she should not use `unwrap()` and `expect()` instead should add the possible _error_ handling scenario. I mean, `expect()` is not for the scenario where you _expect_ an error.  It's the opposite, exactly like `unwrap()`.

src/main.rs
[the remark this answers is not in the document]
snawaz: What is the difference between `context` and `with_context`?  The former ignores `e`?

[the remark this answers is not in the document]
ebkalderon: Both include `e` in the overall chain of error causes, but `with_context()` allows you to examine the contents of `e` as you're formatting the context message, if you so desire.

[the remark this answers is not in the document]
snawaz: I guessed so... after @trha  serial suggestions!

src/main.rs
ebkalderon: I understand, but the point that I'm trying to make is that if `env!("CARGO_PKG_VERSION")` fails, then your application is probably not going to compile at all. The `expect()` message is useful at run-time when debugging broken run-time behavior, sure, but `env!()` is a compile-time invocation. If Cargo is unable to pass in the value of `CARGO_PKG_VERSION` into `rustc` ([source](https://github.com/rust-lang/cargo/blob/d6fa260d1aa74521c7d8c03d3d3c09067a4147d0/src/cargo/core/compiler/compilation.rs#L256)), which must be a semver string, might I add, along with the other required compile-time values, I imagine that your application won't pass the build phase to begin with.

Either way, though, I think it's probably not a good use of our time to bikeshed the usage of `expect()` versus `unwrap()` here in this thread. 😄

snawaz: > The expect() message is useful at run-time when debugging broken run-time behavior, sure, but env!() is a compile-time invocation.

It is called on the `parse()`, not on `env!()`. 

If compiler has bug, then many things are possible. It could also build the project successfully, and it may  fail at the runtime. Like I said, _assertions_ are used to check your _assumptions_. Everytime it fails, you _learn_ new things, new possibility. An _assertion_, when fails, reveals extremely-hard-to-find-bugs somewhere in your tech stack, from your code to the compiler, even in the hardward.

Anyway, I'm okay with going forward and merge it, as I've said in my very first comment, even though we differ on how we think of _assertions_ and _assumptions_.

plumenator: The whole `expect()` vs `unwrap()` argument will become moot after https://github.com/rust-lang/rust/pull/67887 gets release in 1.42

snawaz: Wow. That's an awesome addition, very useful for developers. 

However, `expect` would still make sense because it adds custom message for _users_, not necessarily _developers_. Good messages are helpful for developers as well, though. But that is completely my opinion. :-)

ebkalderon:
Okay, after letting this simmer for a while, I think this PR is good enough to merge. 🚀 Further improvements will be landing in subsequent pull requests.

PR #89, opened 2020-03-02
Don't cross-compile cargo2nix to musl by default

ebkalderon:
### Changed

* Specify `crossSystem ? null` for the `cargo2nix` Rust crate.

Since the `cargo2nix` crate cannot even compile under `musl` on Linux anyway, thanks to the `nix-prefetch-git` dependency added in #62, it no longer makes sense to recommend cross-compiling to that platform by default. Instead, we should default to compiling for the system and `libc` of the build platform, where the command is likely to be running anyway.

This is not a breaking change for downstream users of the `cargo2nix` overlay. It also should not affect future installations of the `cargo2nix` CLI tool because the README already recommended installing with `--arg crossSystem null` anyway.

Fixes #88.

PR #90, opened 2020-03-02
Switch to new Cachix in CI

ebkalderon:
### Changed

* Switch to new `cargo2nix` cache in CircleCI. This change is coupled with the addition of `CACHIX_SIGNING_KEY` to the build-time secrets.
* Push `nix-build` output inside CI build step, using [this configuration](https://github.com/tweag/clodl/blob/master/.circleci/config.yml) as reference.

Should hopefully fix CI for #89.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,8 @@ jobs:
           name: Build
           command: |
             USER=root cachix use cargo2nix
-            nix-build -A ci --show-trace | cachix push cargo2nix
+            cachix push cargo2nix -w &
+            nix-build -A ci --show-trace
           no_output_timeout: 5h
       - save_cache:
           key: nix-store
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,8 +12,8 @@ jobs:
           name: Install cachix
           command: |
             nix-env -iA cachix -f https://cachix.org/api/v1/install
-            USER=root cachix use tenx-cargo2nix
-            USER=root cachix push tenx-cargo2nix -w &
+            USER=root cachix use cargo2nix
+            USER=root cachix push cargo2nix -w &
       - run:
           name: Build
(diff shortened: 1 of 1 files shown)


ebkalderon:
I think I'll build the `ci` attribute locally with `musl`, populate the new Cachix cache, and re-trigger CI here.

ebkalderon:
Local build is almost complete. I will retrigger CI very soon.

ebkalderon:
Okay, I've learned a few things about running Nix inside CircleCI on the free open-source project plan:

1. As described in https://github.com/tenx-tech/cargo2nix/pull/89#issuecomment-593239136, the build jobs often get killed by CircleCI once the Docker image consumes more memory and CPU than the free `medium` instance will allow.
2. You can work around this by building `cargo2nix` locally, pushing the build artifacts yourself up to Cachix, and then re-triggering CI, which will copy the artifacts back down from the cache.
3. `nix-build -A ci | cachix push` only pushes the final build artifact up to the cache. In this case, the resulting NAR from the test runner references none of the tests' build inputs, so a nearly empty derivation gets copied over. The real solution to populate the Cachix cache is to run:
   ```
   nix-build -A rustPkgs.workspace.cargo2nix | cachix push cargo2nix
   ```
   Only then will `nix-build -A ci` in CircleCI complete within a reasonable time and with low enough memory consumption such that it won't be automatically killed.

ebkalderon:
@trha I'm documenting this here just in case we run into these CI issues again. 😄

ebkalderon:
Nice! It looks like Cachix is actually working now, since there are now lines like these appearing in the build output:

```
compressing and pushing /nix/store/llv0s3c99bwjd9y4ibjdizwnyf6aw035-crate-rand_core-0.5.1-x86_64-unknown-linux-musl (220.44 KiB)
```

PR #91, opened 2020-03-06
Add cargo2nix sample projects

### LINKED ISSUE #69: Add some sample projects
I believe we could improve the user experience a bit by providing a few sample Cargo crates and crate workspaces converted to use `cargo2nix`. This would provide useful examples for potential users looking to convert their projects, but might not know exactly how to get started with the instructions given in the README.

Example scenarios:

1. Single crate (`lib` and/or `bin`, combination doesn't exactly matter to `cargo2nix`)
2. Multi-crate workspace
3. Crate with one or more feature flags
4. Crate with extra directories to be included (e.g. a `proto` directory for Protobuf projects)
5. Crate with test runner and some `nativeBuildDependencies`
6. Crate with a custom crate override on a sub-dependency

ebkalderon:
### Added

* Add `examples` directory with a `README.md`.
* Add `1-hello-world` example, showing a simple binary crate being built and tested.
* Add `2-hello-world-with-lib` example, showing a crate with both a `main.rs` and `lib.rs` being built and tested.

### Changed

* Build all examples in CI.

As this is a pretty large PR, I'd recommend going commit-by-commit for the best experience.

Closes #69.


snawaz:
_If you're planning to add multiple sample projects, maybe you can add one at a time. :D_

ebkalderon:
@snawaz I generally agree, but I think I can fit perhaps just one more in this PR, showing a library crate, perhaps. 😁

ebkalderon:
I think this should be ready for review now, so I've marked it accordingly! I've only included two basic examples for the sake of keeping the PR size relatively small.

default.nix
```diff
@@ -77,7 +77,15 @@ rec {
...
     nativeBuildInputs = with rustPkgs; [ cargo rustc ];
   };
+  examples =
+    let
+      importExprsInDir = with pkgs.lib; dir:
+        mapAttrsToList (name: _: import (dir + "/${name}") {})
+          (pkgs.lib.filterAttrs (name: kind: kind == "directory")
+            (builtins.readDir dir));
+    in
+      map (x: x.package) (importExprsInDir ./examples);
```
snawaz: Very nice! Learnt 3 simple Nix functions. 😎

examples/1-hello-world/README.md
```diff
@@ -0,0 +1,134 @@
...
+5. Write the `default.nix`, and optionally a `shell.nix`.
+
+[official instructions]: ../../README.md#install
+
+## Building
+
+To compile the `hello-world` binary with Nix, run:
+
+```bash
+nix-build -A package
```
antigravityla: recommend `my-package` to cue users that this identifier is bound, not a keyword

examples/1-hello-world/README.md
```diff
@@ -0,0 +1,134 @@
...
+$ ./result/bin/hello-world
+Hello, world!
+```
+
+## Running tests
+
+To build and run the test binaries for this project, run:
+
+```bash
+nix-build -A ci
```
antigravityla: Is the user being asked to create this project and then compare their result or just to consume our result?  Going through the motions themselves is likely quite valuable.  At the end say, _The example output of this flow is over in `/sample/`_

It's hard to understand a c2n project without going through the bootstrap IMO.   `ci` just exists out of nowhere at this point.

Okay, I see.  Earlier the instruction says:

> write the default.nix

https://knowyourmeme.com/memes/how-to-draw-an-owl  :grinning: 

Adding a stripped down default.nix (for some reason commentary makes the c2n one scary at first glance) will both fit into the instructions succinctly and lead the user to understand where these build attributes are showing up from

examples/1-hello-world/default.nix
```diff
@@ -0,0 +1,43 @@
+{
+  nixpkgs ? builtins.fetchTarball {
```
antigravityla: Recommend commenting that these are just pins (looks hairy for a hello world) or factoring out to pins.nix or something.

examples/1-hello-world/default.nix
```diff
@@ -0,0 +1,43 @@
...
+  },
+  nixpkgsMozilla ? builtins.fetchTarball {
+    url = https://github.com/mozilla/nixpkgs-mozilla/archive/50bae918794d3c283aeb335b209efd71e75e3954.tar.gz;
+    sha256 = "07b7hgq5awhddcii88y43d38lncqq9c8b2px4p93r5l7z0phv89d";
+  },
+  cargo2nix ? builtins.fetchTarball {
+    url = https://github.com/tenx-tech/cargo2nix/archive/v0.8.0.tar.gz;
+    sha256 = "1qmfijg6l4v3wrlhppv1ayllxlhag7y97xiyh9ihvgjsz4pbcb7b";
+  },
+  system ? builtins.currentSystem,
```
antigravityla: Recommend commenting that these are formal args and their defaults.  Help the user know which parts they need to write and what they need to copy & paste

examples/1-hello-world/default.nix
```diff
@@ -0,0 +1,43 @@
...
+  overlays ? [ ],
+  crossSystem ? null,
+}:
+let
+  pkgs = import nixpkgs {
+    inherit system crossSystem;
+    overlays =
```
antigravityla: You can simplify this by skipping the bindings to `rustOverlay` and `cargo2nixOverlay` and deleting the formal `overlays` arg.  It leaves behind a literal such as

```
overlays = [ (import "${nixpkgsMozilla}/rust-overlay.nix") (import...]
```

Making the code compltely devoid of indirection is super important for hello world.  Literals are direct.

examples/1-hello-world/default.nix
```diff
@@ -0,0 +1,43 @@
...
+  cargo2nix ? builtins.fetchTarball {
+    url = https://github.com/tenx-tech/cargo2nix/archive/v0.8.0.tar.gz;
+    sha256 = "1qmfijg6l4v3wrlhppv1ayllxlhag7y97xiyh9ihvgjsz4pbcb7b";
+  },
+  system ? builtins.currentSystem,
+  overlays ? [ ],
+  crossSystem ? null,
+}:
+let
+  pkgs = import nixpkgs {
```
antigravityla: I would comment that this declaration / binding of pkgs is an overlay of `nixpkgs` and several layers of modifications.  Can't think of a good way to say it off the top of my head.

Key point is to indicate that this is a copy-paste hunk, not something the user wants to dive into at first.

ebkalderon:
@antigravityla Love these suggestions! I'll restructure the projects to be more incremental and understandable in a subsequent PR, for sure. Thanks a ton for the input! ❤️

examples/1-hello-world/default.nix
ebkalderon: That might be a good idea, sure. Maybe we could use a common tool like Niv to simplify the `default.nix`, perhaps?

examples/1-hello-world/README.md
ebkalderon: Would you mind elaborating on what you mean by this, exactly? `package` isn't a language keyword, and I imagine it's pretty self-explanatory what `-A package` does since we assume basic high-level knowledge of Nix for incoming users.

examples/1-hello-world/default.nix
ebkalderon: I see what you mean, but I can't think of any way to simplify this other than leaving a comment. Perhaps you might have another idea? I like the simplicity of this demonstration of [nixpkgs-mozilla](https://github.com/mozilla/nixpkgs-mozilla#rust-overlay) in the README, for example. Maybe we could do something similar?

ebkalderon:
Right, I think I'll merge this PR in as-is, but I'll restructure it to match all of your suggestions. Thanks a lot for the input, everyone!

PR #93, opened 2020-03-17
Simplify and clarify example projects

ebkalderon:
### Changed

* Restructure `hello-world` project to be much simpler and friendlier.
* Restructure second example project to be a simple HTTP client with dependencies on `reqwest` and `tokio`.
* Add `<nixpkgs>` include for examples in CircleCI.
* Reference the example projects in `README.md`.

Sorry for the fairly busy pull request, guys! I took to heart the suggestions made in #91 previously and greatly improved the example projects, making it less of a showcase and more a set of sequential tutorials, with each one building upon concepts established in the last. I tried to make it as user-friendly as possible, and I've updated the README to point directly to the `examples` folder as well.

Like before, I'd recommend following along commit-by-commit. If you'd like, maybe try reading the `1-hello-world` and `2-bigger-project` READMEs in rendered mode to get a feel of how the documents will look in practice.

Follow-up to #91.


ebkalderon:
Okay, I fixed up the most egregious of typos and Markdown missteps, haha. Should be ready for review!

examples/1-hello-world/README.md
```diff
@@ -4,131 +4,283 @@ This is very simple `bin` crate generated by `cargo new`, with no additional
...
+
+  pkgs = import <nixpkgs> {
+    inherit system;
+    overlays = [ rustOverlay cargo2nixOverlay ];
+  };
+
+  rustPkgs = pkgs.rustBuilder.makePackageSet' {
+    rustChannel = "stable";
+    packageFun = import ./Cargo.nix;
+    packageOverrides = pkgs: pkgs.rustBuilder.overrides.all;
```
trha: Same here. Should we omit this line and reintroduce it in the second example?

examples/1-hello-world/README.md
```diff
@@ -4,131 +4,283 @@ This is very simple `bin` crate generated by `cargo new`, with no additional
...
+
+There's a bit to unpack in the changes shown above. We use `makePackageSet'` to
+import our `Cargo.nix` file and build it using the `"stable"` release channel of
+Rust. The output of this function is a tree structure containing the entire
+dependency graph from our `Cargo.nix`, which we call `rustPkgs`.
+
+The structure of `rustPkgs` is organized as follows:
+
+```nix
+             rustPkgs."unknown"."0.1.0".hello-world { }
```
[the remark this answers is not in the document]
ebkalderon: D'oh! I totally messed this one up. I'll fix it, including the ASCII art. 🙃

examples/1-hello-world/README.md
```diff
@@ -4,131 +4,283 @@ This is very simple `bin` crate generated by `cargo new`, with no additional
...
+need to have a `default.nix` present at the project root. Let's create one!
+
+To start, let's create a new file called [`default.nix`] and declare a function
+with the following arguments:
+
+[`default.nix`]: ./default.nix
+
+```nix
+{
+  system ? builtins.currentSystem,
```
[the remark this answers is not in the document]
ebkalderon: I personally don't mind having `system` present, as it seems to be a common pattern for many Nix expressions to provide `system` as an argument. I do totally agree with leaving out `crossSystem`, though, which is why I decided to ignore its existence for these two examples.

examples/2-bigger-project/README.md
```diff
@@ -0,0 +1,224 @@
+# A bigger project
+
+This is larger `bin` project which requires some dependencies from Crates.io and
```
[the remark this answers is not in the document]
ebkalderon: Nice catch! ❤️

examples/2-bigger-project/README.md
```diff
@@ -0,0 +1,224 @@
...
+is compiled using the latest release of Rust.
+
+## Introduction
+
+This example assumes that you have already read through the [Hello World]
+project and have your machine set up with Nix and `cargo2nix`. If you this is
```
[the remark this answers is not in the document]
ebkalderon: Whoops, thanks for catching that!

examples/1-hello-world/README.md
ebkalderon: Honestly, I'm not sure. On one hand, we could remove it since it's technically optional and everything will still build without it.

But on the other hand, if we choose to remove it, I think it should be made clear that it will default to `pkgs.rustBuilder.overrides.all` regardless of whether it's specified or not, and we don't need to add it in order to get `reqwest` to compile. If it's present from the very beginning, then the user knows we didn't need to do anything special in order to get their system dependencies built.

ebkalderon: I don't know, what do you think @trha? 😄

ebkalderon:
@antigravityla Those are some good points! Would you like me to revise the wording in the `README`s a bit? Does it need to be made more concise?

examples/1-hello-world/README.md
antigravityla: The first thing to get across in a hello world is what you absolutely can't run without.  Recommend omitting where possible

ebkalderon: Okay, that's fair! Maybe I could remove it from `hello-world`, but perhaps leave it in `bigger-project` but commented out, with a note saying "this is where the crate overrides come from."

ebkalderon: @antigravityla Okay, that's fair! Maybe I could remove it from `hello-world`, but perhaps leave it in `bigger-project` but commented out, with a note saying "this is where the crate overrides come from?"

ebkalderon:
@antigravityla I cleaned up a few bits in both examples that I thought read a bit redundantly. Hope this makes the project-building parts a bit less repetitious!

examples/1-hello-world/README.md
ebkalderon: Done in d214c36! I decided to remove it from `hello-world` but introduce it commented out in `bigger-project`.

examples/1-hello-world/README.md
```diff
@@ -135,9 +132,7 @@ with the following arguments:
 }:
 ```
 
-As shown above, we declare this function with three arguments: `system`,
```
antigravityla: Exactly.  Reader is going to see the arguments enumerated right below here, so no need to list twice.

ebkalderon: Yup, exactly what I was thinking when I read it over again. 😁

examples/1-hello-world/README.md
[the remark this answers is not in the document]
ebkalderon: I also forgot to mention that we actually link to a [Nix pill](https://nixos.org/nixos/nix-pills/nixpkgs-parameters.html#idm140737319760528) in the document when talking about `system` being commonplace in Nix expressions. 😄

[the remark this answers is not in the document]
trha: Ah okay. Let's leave `system` there as it doesn't cause any confusion and was explained clearly in the `README`.

examples/2-bigger-project/README.md
```diff
@@ -147,7 +147,7 @@ let
   rustPkgs = pkgs.rustBuilder.makePackageSet' {
     rustChannel = "stable";
     packageFun = import ./Cargo.nix;
-    packageOverrides = pkgs: pkgs.rustBuilder.overrides.all;
+    # packageOverrides = pkgs: pkgs.rustBuilder.overrides.all; # Implied, if unspecified
```
trha: Unfortunately the default value for `packageOverrides` is an empty list. I wanted to make it explicit so when a native dependency is missing, users know where to look for. Do you think we should make `all` the default instead?

ebkalderon: Ahh, I see what you mean. I guess we do have to specify it instead. I think we should have it default to `all`, though, since that improves the user experience considerably. 👍

Nix is lazy, so it's not like it will attempt to build any of it unless it needs to. It also doesn't increase the fetch size of the overlay since we're already cloning the entire `cargo2nix` repository with `fetchTarball`/`fetchGit` anyway.

If the user wants to build with absolutely no default overrides, which I imagine a select few users might, they could provide:

```
packageOverrides = _: [];
```

Perhaps we could merge this PR first and then create a follow-up PR changing the default value?

trha: Yeah let's do that!

PR #94, opened 2020-03-17
Make packageOverrides default to overrides.all

ebkalderon:
### Changed

* Make `packageOverrides` default to `rustBuilder.overrides.all`.

Follow-up to https://github.com/tenx-tech/cargo2nix/pull/93#pullrequestreview-375921456.

### DIFF
--- a/overlay/make-package-set/simplified.nix
+++ b/overlay/make-package-set/simplified.nix
@@ -7,7 +7,7 @@
 args@{
   rustChannel,
   packageFun,
-  packageOverrides ? _: rustBuilder.overrides.all,
+  packageOverrides ? pkgs: pkgs.rustBuilder.overrides.all,
   ...
 }:
 let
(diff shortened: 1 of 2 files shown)


overlay/make-package-set/simplified.nix
```diff
@@ -7,7 +7,7 @@
 args@{
   rustChannel,
   packageFun,
-  packageOverrides ? _: [ ],
+  packageOverrides ? _: rustBuilder.overrides.all,
```
trha: Let's use the overrides from the `pkgs` argument here. This `packageOverrides` is different from the one in `full.nix`, as it's a function and is called twice, once with `pkgs` and once with `buildPackages`.

```suggestion
  packageOverrides ? pkgs: pkgs.rustBuilder.overrides.all,
```

ebkalderon: Love it! Let's do that.

PR #95, opened 2020-03-17
Prepare release 0.8.2

ebkalderon:
### Changed

* Bump `cargo2nix` crate version to 0.8.2 and regenerate `Cargo.nix`.
* Update example projects to fetch overlay from future tag `v0.8.2`.

Given the number of changes present in `master`, as well as the `packageOverride` change from #94, we should prepare a new release.


ebkalderon:
The example projects' READMEs point to `v0.8.2` currently, but the `default.nix` files themselves are pulling in `v0.8.1` for the moment until the release is actually tagged. 😄 A subsequent pull request immediately after this one will update their `default.nix` files to match the documentation.

ebkalderon:
@trha Haven't compiled the changelog just yet. 😆 I'll have to do that just before I tag the release.

However, there seems to be a spurious issue in the latest commit with `lazy-static` not being able to compile due to "unnecessary braces around return value," which seems to be a warning and we have `--cap-lints warn` set, IIRC. Not sure why CI was failing on that.

ebkalderon:
Worst case, we can force-merge this and resolve the issue later once we tag the new release.

ebkalderon:
Ahh, I forgot to restore [this line](https://github.com/tenx-tech/cargo2nix/pull/95/files#diff-4cbeb0d76ed08afa6975ad65f441b828R24). It's no problem, then. Let's drop the latest commit, force-merge this in, and tag the new release so CI can resolve itself.

ebkalderon:
@trha Re-requesting your review!

PR #98, opened 2020-03-19
Attempt to copy tempfile if atomic rename fails

### LINKED ISSUE #97: Cannot write Cargo.nix directly
I installed the latest version (0.8.2) of `cargo2nix` yesterday using `nix-env -iA package -f https://github.com/tenx-tech/cargo2nix/tarball/master`.

I cannot create `Cargo.nix` directly using `cargo2nix -f`. It panics with `Invalid cross-device link (os error 18)`:

```
$ cargo2nix -f
error: could not write file to Cargo.nix

Caused by:
  failed to persist temporary file: Invalid cross-device link (os error 18)
```

As a workaround, `cargo2nix -s > Cargo.nix` worked for me.

I'm using Linux on a 5.5.9 Kernel using ext4 as filesystem.

ebkalderon:
### Fixed

* Attempt to copy tempfile if atomic rename fails.

This should fix cases where `tempfile::NamedTempFile::persist()` fails due to `Invalid cross-device link (os error 18)`, which is likely due to atomic file moves being problematic on Linux with `/tmp` on `tmpfs` or an otherwise separate filesystem from the root partition.

Would you mind trying this branch and confirm whether it fixes your issue, @vbrandl?

Fixes #97.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -163,9 +163,11 @@ fn write_to_file(file: impl AsRef<Path>) -> Result<()> {
 
     generate_cargo_nix(&mut temp_file)?;
 
-    temp_file
-        .persist(path)
-        .context(format!("could not write file to {}", path.display()))?;
+    if let Err(err) = temp_file.persist(path) {
+        let (_, temp_path) = err.file.keep()?;
+        std::fs::copy(temp_path, path)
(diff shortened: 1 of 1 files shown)


ebkalderon:
Rebased against the latest `master` after merging #99. Would you please re-review, @trha?

PR #104, opened 2020-03-20
Remove pkg-config nativeBuildInput in mkCrate

ebkalderon:
### Added

* Add `buildPackages.pkg-config` as propagated native build input to `pkg-config` override.

### Removed

* Remove `pkg-config` native build input for `mkCrate`.

I don't really think it makes sense to have `pkg-config` as a required native build input for `mkCrate` but not any other meta-packaging or meta-compilation tool, for example. Instead, we should add `pkg-config` to the `pkg-config` crate override in `overlay/overrides.nix`. This way, crates that require it will have access to it, while all others that don't will avoid pulling it in unnecessarily.

I've tried rebuilding `cargo2nix` itself (which depends on `pkg-config` for a few dependencies) as well as larger crates such as `rust-analyzer` without any apparent issues. I'd appreciate a second or third pair of eyes to confirm that this change doesn't break anything. 😄

### DIFF
--- a/overlay/mkcrate.nix
+++ b/overlay/mkcrate.nix
@@ -97,7 +97,7 @@ let
     inherit src version meta;
     buildInputs = runtimeDependencies;
     propagatedBuildInputs = concatMap (drv: drv.propagatedBuildInputs) runtimeDependencies;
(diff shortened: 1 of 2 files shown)


ebkalderon:
Just rebased on the latest `master` after merging #105.

PR #109, opened 2020-04-25
Update fetchGit

### LINKED ISSUE #108: Git Deps Support Broken
## First the Broken Parts

The overlay, necessary for consuming generated `Cargo.nix` currently contains broken references to `fetchgitPrivate`
https://github.com/tenx-tech/cargo2nix/blob/master/overlay/lib/fetch.nix#L11-L12

```shell
$ nix-channel --list
...
nixos https://nixos.org/channels/nixos-20.03

$ nix repl '<nixpkgs>'
nix-repl> buildPackages.fetchgitPrivate
error: attribute 'fetchgitPrivate' missing, at (string):1:1
```
`fetchgitPrivate` is actually [dead in master nixpkgs](https://github.com/NixOS/nixpkgs/commit/e578b8499f9e49441d9761875659381f95975fee)

## Crafting a Fix

There are three contexts to reconcile:

### Cargo
[Specifying dependencies from git repositories](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories)
> Since we haven’t specified any other information, Cargo assumes that we intend to use the latest commit on the master branch to build our package. You can combine the git key with the *rev*, *tag*, or *branch* keys to specify something else.

Cargo lockfiles do not appear to include checksums for git dependencies.

### Cargo2Nix
https://github.com/tenx-tech/cargo2nix/blob/master/templates/Cargo.nix.tera#L49-L55

### Nix
`fetchgit` doesn't fetch at evaluation time
`fetchGit` doesn't accept a sha256 argument (this is being [calculated by cargo2nix](https://github.com/tenx-tech/cargo2nix/blob/master/src/main.rs#L553-L557) since it's not in the cargo docs.)
`fetchFromGitHub` will of course only work for Github

The most functionality will be covered by switching `fetchGit`.   The downside is that the sha256 argument needs to be ignored and no integrity check will be performed.  This sha256 pin & check may be possible for generic git, but the appropriate place for a fix would be upstream in nixpkgs, not in cargo2nix.

## Tracking Integrity Issues

I attempted to contextualize the choices within the Rust community and found Boats discussing [the update framework](https://theupdateframework.io/) https://github.com/rust-lang/cargo/issues/4768  I was unable to find an issue to add sha256 support to git fetchers generically in nixpkgs. Recommend adding tracking issues to this repo if other integrity check work is known.

macs-overdrv:
fetchgitPrivate was deprecated in nixpkgs master.

This change drops support for sha256 integrity checking for git dependencies.
The unused sha256 is noted in the Cargo.nix generation template.  The integrity
check is only relevant to fetchCrateGit.

* Template change was tested by re-generating a Cargo.nix with git deps
* Overlay change was tested by patching the overlay to build a project with git deps
* Changes to `fetchCrateAlternativeRegistryExpensive` have not been tested but are broken as is

Fixes #108

### DIFF
--- a/overlay/lib/fetch.nix
+++ b/overlay/lib/fetch.nix
@@ -8,10 +8,10 @@ rec {
 
   fetchCrateGit = { url, name, version, rev, sha256 }:
     let
-      inherit (buildPackages) runCommand jq remarshal fetchgitPrivate;
-      repo = fetchgitPrivate {
+      inherit (buildPackages) runCommand jq remarshal;
+      repo = builtins.fetchGit {
         name = "${name}-${version}-src";
-        inherit url rev sha256;
+        inherit url rev;
       };
     in
       /. + builtins.readFile (runCommand "find-crate-${name}-${version}"
@@ -35,7 +35,7 @@ rec {
   fetchCrateAlternativeRegistryExpensive = { index, name, version, sha256 }: with buildPackages; stdenvNoCC.mkDerivation {
     name = "${name}-${version}.tar.gz";
 
-    inherit (fetchgitPrivate { url = git://dummy; }) GIT_SSH SSH_AUTH_SOCK;
+    inherit (builtins.fetchGit { url = git://dummy; }) GIT_SSH SSH_AUTH_SOCK;
     INDEX = index;
     CRATE_NAME = name;
     CRATE_VERSION = version;
--- a/templates/Cargo.nix.tera
+++ b/templates/Cargo.nix.tera
@@ -51,7 +51,7 @@ in
       name = "{{ crate.name }}";
       version = "{{ crate.version }}";
       rev = "{{ crate.source.Git.rev }}";
-      sha256 = "{{ crate.source.Git.sha256 }}";
+      sha256 = "{{ crate.source.Git.sha256 }}"; {# unused #}
     };
     {%- elif crate.source.Local.path %}
     src = fetchCrateLocal ./{{ crate.source.Local.path }};


overlay/lib/fetch.nix
```diff
@@ -35,7 +35,7 @@ rec {
   fetchCrateAlternativeRegistryExpensive = { index, name, version, sha256 }: with buildPackages; stdenvNoCC.mkDerivation {
     name = "${name}-${version}.tar.gz";
 
-    inherit (fetchgitPrivate { url = git://dummy; }) GIT_SSH SSH_AUTH_SOCK;
+    inherit (builtins.fetchGit { url = git://dummy; }) GIT_SSH SSH_AUTH_SOCK;
```
ebkalderon, line 38: This expression won't actually evaluate since it's done at evaluation time, unfortunately. We need to either inherit `GIT_SSH` and `SSH_AUTH_SOCK` explicitly using `builtins.getEnv` or rewrite `fetchCrateAlternativeRegistryExpensive` to fetch the registry index using `builtins.fetchGit` properly. CC @trha

ebkalderon: This expression won't evaluate successfully since it's done at evaluation time, and `builtins.fetchGit` doesn't have `GIT_SSH` nor `SSH_AUTH_SOCK` attributes, AFAIK.

We need to either inherit `GIT_SSH` and `SSH_AUTH_SOCK` explicitly using `builtins.getEnv` or rewrite `fetchCrateAlternativeRegistryExpensive` to fetch the registry index using `builtins.fetchGit` properly. CC @trha

ebkalderon, line 38: This expression won't evaluate successfully since it's done at evaluation time, and `builtins.fetchGit` doesn't have `GIT_SSH` nor `SSH_AUTH_SOCK` attributes, AFAIK.

We need to either load `GIT_SSH` and `SSH_AUTH_SOCK` explicitly using `builtins.getEnv` or rewrite `fetchCrateAlternativeRegistryExpensive` to fetch the registry index using `builtins.fetchGit` properly. CC @trha

ebkalderon: @macs-overdrv I think for the sake of merging this PR quickly to unblock 99% of users, we should just use `builtins.getEnv` to restore the current behavior for alternative registries.

```suggestion
    GIT_SSH = builtins.getEnv "GIT_SSH";
    SSH_AUTH_SOCK = builtins.getEnv "SSH_AUTH_SOCK";
```

We can work on switching to `builtins.fetchGit` for alternative registries at a later point, as that is a less frequently used feature.

Mic92: By the way this cannot be used in nixpkgs.

ebkalderon: @Mic92 Thanks for the update! Is it available in the latest stable Nixpkgs, at least?

Mic92: It's in nix that we use in nixpkgs in 20.03.

ebkalderon: Gotcha, thanks. Is this a temporary or permanent measure? Could you post a link to a PR or tracking issue regarding this restriction for future readers?

Mic92: Those builtins fetchers are not allowed when using `restrict-eval`. Which is what we use when evaluating nixpkgs: https://github.com/NixOS/ofborg/#how-does-ofborg-call-nix-build

I think this is needed to not depend on network for evaluating and to avoid import from derivation.

```
$ nix eval --option restrict-eval true '(builtins.fetchGit { url = "https://github.com/NixOS/nix"; rev = "ecd4e52a580db0b928b84c049e72472c00f788db";  })'
warning: unknown setting 'experimental-features'
error: access to URI 'https://github.com/NixOS/nix' is forbidden in restricted mode
```

ebkalderon: Ah, I see what you mean. I think it's fine for fetching private dependencies from alternative registries, since those will never make their way into Nixpkgs.

However, in the case of plain Git dependencies, we cannot use `fetchgit` either, since that doesn't work with fetching from private GitHub repos. We used to use `fetchgitPrivate` for this, but since that has been replaced with `builtins.fetchGit`, this is our only option left. Does there exist an alternative path?

Mic92: I guess it should be fine than.

ebkalderon: That is indeed valuable to know. Thank you! I wonder if maybe there's a heuristic we could apply which defaults to `fetchgit` and only switches to `builtins.fetchGit` if absolutely necessary? 🤔

ebkalderon:
Considering that this pull request has been open for about a month and has grown out of date with `master`, and we need this fix to be merged in sooner rather than later, I think we should cherry pick this commit into a new pull request which includes the necessary alternative registry fixes and a notice in the README regarding the caveats of `builtins.fetchGit`.

Thank you so much for reporting the issue and doing the brunt of the work, @macs-overdrv! ❤️

ebkalderon:
Closed in favor of #113.

PR #113, opened 2020-05-15
Update to builtins.fetchGit

### LINKED ISSUE #108: Git Deps Support Broken

ebkalderon:
### Added

* Add notice regarding caveats of `builtins.fetchGit` to `README.md`.
### Changed

* Switch from `fetchgitPrivate` (which was removed upstream) to `builtins.fetchGit`.
* Ignore the `sha256` argument used in `fetchCrateGit` for backwards compatibility. We should remove this argument from the overlay and codegen completely in a subsequent minor release.

Original commit authored by @macs-overdrv and cherry-picked from #109. CC @Mic92

Fixes #108.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,11 @@ runHook runCargo
    error: while parsing a TOML string at /nix/store/.../overlay/mkcrate.nix:31:14: Bare key 'cfg(target_os = "linux")' cannot contain whitespace at line 45
    ```
 
+3. Git dependencies and crates from alternative Cargo registries rely on
+   `builtins.fetchGit` to support fetching from private Git repositories. This
+   means that such dependencies cannot be evaluated with `restrict-eval`
+   applied.
+
 ## Design
 
 This Nixpkgs overlay builds your Rust crates and binaries by first pulling the
--- a/README.md
+++ b/README.md
@@ -92,6 +92,7 @@ runHook runCargo
 1. Many `crates.io` public crates may not build using the current Rust compiler,
    unless a lint cap is put on these crates. For instance, `cargo2nix` caps all
    lints to `warn` by default.
+
 2. Nix 2.1.3 ships with a broken `builtins.fromTOML` function which is unable to
    parse lines of TOML that look like this:
 
@@ -113,6 +114,10 @@ runHook runCargo
    `builtins.fetchGit` to support fetching from private Git repositories. This
    means that such dependencies cannot be evaluated with `restrict-eval`
    applied.
+
+   Also, if your Git dependency is tied to a Git branch, e.g. `master`, and you
+   would like to force it to update on upstream changes, you should append
+   `--option tarball-ttl 0` to your `nix-build` command.
 
 ## Design
 
--- a/overlay/lib/fetch.nix
+++ b/overlay/lib/fetch.nix
@@ -34,9 +34,8 @@ rec {
   # to be downloaded, the whole index needs to be recloned to get the download URL, which is potentially expensive.
   fetchCrateAlternativeRegistryExpensive = { index, name, version, sha256 }: with buildPackages; stdenvNoCC.mkDerivation {
     name = "${name}-${version}.tar.gz";
+    src = builtins.fetchGit { url = index; ref = "master"; };
 
-    inherit (builtins.fetchGit { url = git://dummy; }) GIT_SSH SSH_AUTH_SOCK;
-    INDEX = index;
     CRATE_NAME = name;
     CRATE_VERSION = version;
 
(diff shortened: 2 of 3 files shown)


overlay/lib/fetch.nix
```diff
@@ -34,22 +34,20 @@ rec {
   # to be downloaded, the whole index needs to be recloned to get the download URL, which is potentially expensive.
   fetchCrateAlternativeRegistryExpensive = { index, name, version, sha256 }: with buildPackages; stdenvNoCC.mkDerivation {
     name = "${name}-${version}.tar.gz";
+    src = builtins.fetchGit { url = index; ref = "master"; };
```
antigravityla, line 37: In the original PR conversation:

https://github.com/tenx-tech/cargo2nix/pull/109#discussion_r421936521

If this suggestion is true, I would expect to see some treatment of the ssh env carried forward. Instead all mentions of ssh env are deleted.

An aside, the `GIT_SSH` being deleted below looks like a rather crazed hack;  it was originally getting environment variables into scope for build time use of `git` further down by failing a `fetchGit`?  :scream:

README.md
```diff
@@ -109,6 +109,11 @@ runHook runCargo
    error: while parsing a TOML string at /nix/store/.../overlay/mkcrate.nix:31:14: Bare key 'cfg(target_os = "linux")' cannot contain whitespace at line 45
    ```
 
+3. Git dependencies and crates from alternative Cargo registries rely on
+   `builtins.fetchGit` to support fetching from private Git repositories. This
+   means that such dependencies cannot be evaluated with `restrict-eval`
```
antigravityla, line 114: Suggest mentioning the `ttl` if that's also still a thing

ebkalderon: Good idea. I should add that caveat in specifically to users of alternative Cargo registries, to which I imagine this would be the most applicable.

overlay/lib/fetch.nix
ebkalderon: There should be no need to load `$GIT_SSH` and `$SSH_AUTH_SOCK` explicitly under this new method because those were "crazed hacks" normalized by Nixpkgs and specific to the now-removed [`fetchgitPrivate`](https://github.com/NixOS/nixpkgs/blob/65e7d9c8c94d934c494ab4486992112db6237dc5/pkgs/build-support/fetchgit/private.nix). `$GIT_SSH` used to point to the Nix store path of the `ssh` binary from `openssh`, while `$SSH_AUTH_SOCK` used to be defined in terms of the `ssh-auth-sock=` setting specified by the user in their `$NIX_PATH`.

With `builtins.fetchGit`, this is no longer necessary. Because this command runs at evaluation-time rather than at build-time, SSH credentials required for private repos are automatically inherited from the environment without needing any such hacks.

This also means that we can also remove the `socat`hacks in CI (and possibly the macOS `nix-docker` remote builder), since all the authorization and cloning of private repositories happens outside the Nix builder.

README.md
ebkalderon: Done in the latest commit!

ebkalderon:
@eupn @antigravityla I've added an additional notice to the README regarding `tarball-ttl` and its applicability to `builtins.fetchGit`. Please re-review at your leisure!

overlay/lib/fetch.nix
antigravityla: > If this suggestion is true, I would expect to see some treatment of the ssh env carried forward. Instead all mentions of ssh env are deleted.

Sounds like it wasn't true.  I buy all of this.  It's how I understood it.

ebkalderon: Indeed it wasn't. I experimented a bit locally, and I also checked the original `fetchgitPrivate` source, and found it wasn't the case. 😄

ebkalderon:
Okay, I think this should be good to merge! Thanks for the reviews, everyone.

PR #114, opened 2020-05-26
Fix pkg-config override, indirectly fixing curl-sys

ebkalderon:
### Changed

* Use `pkgs.pkg-config` over `pkgs.buildPackages.pkg-config`.
* Pass `pkgs.pkg-config` to `propagatedBuildInputs` instead of `propagatedNativeBuildInputs`.

### Fixed

* Always include `pkgs.pkg-config` as a propagated build input in the crate override, regardless of whether we are cross-compiling or not, fixing a bug introduced in #104.

### DIFF
--- a/overlay/overrides.nix
+++ b/overlay/overrides.nix
@@ -127,11 +127,11 @@ in rec {
     overrideAttrs = drv: {
       propagatedBuildInputs = drv.propagatedBuildInputs or [ ] ++ [
         pkgs.pkg-config
-      ] ++ lib.optional (pkgs.stdenv.hostPlatform != pkgs.stdenv.buildPlatform) [
+      ] ++ lib.optional (pkgs.stdenv.hostPlatform != pkgs.stdenv.buildPlatform)
         (propagateEnv "pkg-config" [
           { name = "PKG_CONFIG_ALLOW_CROSS"; value = "1"; }
         ])
-      ];
+      ;
     };
   };
 
--- a/overlay/overrides.nix
+++ b/overlay/overrides.nix
@@ -127,11 +127,10 @@ in rec {
     overrideAttrs = drv: {
       propagatedBuildInputs = drv.propagatedBuildInputs or [ ] ++ [
         pkgs.pkg-config
-      ] ++ lib.optional (pkgs.stdenv.hostPlatform != pkgs.stdenv.buildPlatform)
         (propagateEnv "pkg-config" [
           { name = "PKG_CONFIG_ALLOW_CROSS"; value = "1"; }
         ])
-      ;
+      ];
     };
   };
 
--- a/overlay/overrides.nix
+++ b/overlay/overrides.nix
@@ -130,8 +130,6 @@ in rec {
           (propagateEnv "pkg-config" [
             { name = "PKG_CONFIG_ALLOW_CROSS"; value = "1"; }
           ])
-        ];
-        propagatedNativeBuildInputs = drv.propagatedNativeBuildInputs or [ ] ++ [
           pkgs.buildPackages.pkg-config
         ];
       };
--- a/overlay/overrides.nix
+++ b/overlay/overrides.nix
@@ -122,19 +122,18 @@ in rec {
     };
   };
 
-  pkg-config = if pkgs.stdenv.hostPlatform != pkgs.stdenv.buildPlatform
-    then makeOverride {
-      name = "pkg-config";
-      overrideAttrs = drv: {
-        propagatedBuildInputs = drv.propagatedBuildInputs or [ ] ++ [
-          (propagateEnv "pkg-config" [
-            { name = "PKG_CONFIG_ALLOW_CROSS"; value = "1"; }
-          ])
-          pkgs.buildPackages.pkg-config
-        ];
-      };
-    }
-    else nullOverride;
+  pkg-config = makeOverride {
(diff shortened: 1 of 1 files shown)


overlay/overrides.nix
```diff
@@ -84,7 +84,12 @@ in rec {
   curl-sys = makeOverride {
     name = "curl-sys";
     overrideAttrs = drv: {
-      propagatedBuildInputs = drv.propagatedBuildInputs or [ ] ++ [ (patchCurl pkgs) ];
+      propagatedBuildInputs = drv.propagatedBuildInputs or [ ] ++ [
+        (patchCurl pkgs)
+        pkgs.pkg-config
+      ];
+    } // lib.optionalAttrs (pkgs.stdenv.hostPlatform != pkgs.stdenv.buildPlatform) {
+      PKG_CONFIG_ALLOW_CROSS = "1";
```
trha, line 92: Are these lines really necessary? They're just duplicating what we already have for `pkg-config`.

overlay/overrides.nix
```diff
@@ -122,21 +127,18 @@ in rec {
...
+  pkg-config = makeOverride {
+    name = "pkg-config";
+    overrideAttrs = drv: {
+      propagatedBuildInputs = drv.propagatedBuildInputs or [ ] ++ [
+        pkgs.pkg-config
+      ] ++ lib.optional (pkgs.stdenv.hostPlatform != pkgs.stdenv.buildPlatform) [
+        (propagateEnv "pkg-config" [
+          { name = "PKG_CONFIG_ALLOW_CROSS"; value = "1"; }
+        ])
+      ];
```
trha, line 139: We don't need to pass a list to `optional`,  as it already returns a list.
`optional = cond: elem: if cond then [elem] else [];`
```suggestion
      ] ++ lib.optional (pkgs.stdenv.hostPlatform != pkgs.stdenv.buildPlatform)
        (propagateEnv "pkg-config" [
          { name = "PKG_CONFIG_ALLOW_CROSS"; value = "1"; }
        ])
      ;
```

overlay/overrides.nix
ebkalderon: Given that `curl-sys` doesn't rely on the `pkg-config` Rust crate but instead on the system `pkg-config` directly, we need to duplicate these options here as well. Is this not the case?

overlay/overrides.nix
ebkalderon: Good catch, man! I didn't realize this.

overlay/overrides.nix
trha: `curl-sys` [does list](https://docs.rs/crate/curl-sys/0.4.31+curl-7.70.0/source/Cargo.toml) `pkg-config` as a build dependency so the dependencies specified in the override for `pkg-config` should be propagated correctly.

Let's try removing these lines and see if the build still works.

ebkalderon: Sure, I'll try that right now. 👍

ebkalderon: Okay, it looks reasonable to me! I've dropped the old commits and force-pushed. Currently awaiting the build to check if it all works, but either way, thanks for the correction.

ebkalderon: I'll re-request the review as soon as it passes in CI! 🤞

ebkalderon: Works! 🎉 Thanks a lot. Should be ready for re-review now.

overlay/overrides.nix
```diff
@@ -130,8 +130,6 @@ in rec {
           (propagateEnv "pkg-config" [
             { name = "PKG_CONFIG_ALLOW_CROSS"; value = "1"; }
           ])
-        ];
-        propagatedNativeBuildInputs = drv.propagatedNativeBuildInputs or [ ] ++ [
```
antigravityla, line 134: What about `drv.propagatedNativeBuildInputs`?

overlay/overrides.nix
```diff
@@ -127,11 +127,10 @@ in rec {
     overrideAttrs = drv: {
       propagatedBuildInputs = drv.propagatedBuildInputs or [ ] ++ [
         pkgs.pkg-config
-      ] ++ lib.optional (pkgs.stdenv.hostPlatform != pkgs.stdenv.buildPlatform)
```
antigravityla: Even if true... I don't see how removing this switch can make the result _more_ reliable.  The commit note is not exactly a strong reassurance :cat2:

overlay/overrides.nix
ebkalderon: Are you curious about what `drv.propagatedNativeBuildInputs` does? Or why the switch to `drv.propagatedBuildInputs` matters?

overlay/overrides.nix
ebkalderon: The issue is that `pkgs.stdenv.hostPlatform != pkgs.stdenv.buildPlatform` will never succeed because the `pkg-config` crate is a build dependency of the crate being cross-compiled. This means that this environment variable is never set and never propagated up the chain to crates that need it, e.g. `curl-sys`. Since this conditional isn't a reliable way of checking whether we are cross-compiling or not, and neither @trha and I are aware of any functional alternatives that work around this issue, we decided to set this variable always. I can't think of any reason where this would cause an issue.

PR #115, opened 2020-05-30
Remove empty link-flags files

antigravityla:
Background is that when certain programs (home manager) are building a profile
to activate, these ./link-flags files are going to collide even though they
don't end up in the user's profile.  While this may be a buggy home manager
behavior, it would also be better not to have unnecessary outputs.

A proper fix needs to start by asking what these files are used for after
building the current crate and its dependents.  This file does not appear to
serve any standard role outside of the cargo2nix context and some of the
contents may be extraneous.

The derivation for cargo2nix will output a .link-flags with flags inside in an
otherwise empty lib folder while rust-analyzer with a busy lib folder has an
empty .link-flags

With the current home manager behavior, any two cargo2nix projects that leave
behind these identically named files will collide.  Even if empty ./link-flags
and ./link-flags with no rlib associations are removed, two cargo2nix
derivations that link to native libraries and have rlib outputs are enough to
trigger the collision.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -78,7 +78,7 @@ rec {
   # `cargo build` with in the shell should just work.
   shell = pkgs.mkShell {
     inputsFrom = pkgs.lib.mapAttrsToList (_: pkg: pkg { }) rustPkgs.noBuild.workspace;
-    nativeBuildInputs = with rustPkgs; [ cargo rustc ];
+    nativeBuildInputs = with rustPkgs; [ cargo rustc rust-src ];
   };
   examples =
     let
--- a/overlay/make-package-set/full.nix
+++ b/overlay/make-package-set/full.nix
@@ -11,6 +11,7 @@
   packageFun,
   cargo,
   rustc,
+  rust-src,
   buildRustPackages ? null,
   localPatterns ? [ ''^(src|tests)(/.*)?'' ''[^/]*\.(rs|toml)$'' ],
   packageOverrides ? rustBuilder.overrides.all,
@@ -55,7 +56,7 @@ lib.fix' (self:
     });
 
   in packageFunWith { mkRustCrate = mkRustCrate'; buildRustPackages = buildRustPackages'; } // {
-    inherit rustPackages callPackage cargo rustc pkgs;
+    inherit rustPackages callPackage cargo rustc rust-src pkgs;
     noBuild = packageFunWith {
       mkRustCrate = lib.makeOverridable mkRustCrateNoBuild { };
       buildRustPackages = buildRustPackages'.noBuild;
--- a/overlay/make-package-set/simplified.nix
+++ b/overlay/make-package-set/simplified.nix
@@ -14,7 +14,7 @@ let
   rustChannel = buildPackages.rustChannelOf {
     channel = args.rustChannel;
   };
-  inherit (rustChannel) cargo;
+  inherit (rustChannel) cargo rust-src;
   rustc = rustChannel.rust.override {
     targets = [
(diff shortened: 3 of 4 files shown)


antigravityla:
```
$ rg '.link-flags'

overlay/utils.sh
32:        if [ -f "$crate/lib/.link-flags" ]; then
33:            cat $crate/lib/.link-flags
44:        if [ -f "$crate/lib/.link-flags" ]; then
45:            cat $crate/lib/.link-flags
134:                    local link_flags=$out/lib/.link-flags
157:    touch $out/lib/.link-flags
158:    loadExternCrateLinkFlags $dependencies >> $out/lib/.link-flags
159:    if [ ! -s $out/lib/.link-flags ]; then
160:      rm $out/lib/.link-flags

examples/1-hello-world/README.md
266:    └── .link-flags

examples/2-bigger-project/README.md
202:│   └── .link-flags
```

To start some due diligence, other references check for the existence of the file before attempting to read it.  It seems to be anticipated at that time that the flags might not always be written.

The example projects include 1 and 3 include a `lib/.link-flags` file with no contents.  I wasn't able to check the 2nd project's output since it failed to build on Linux, a separate matter.

antigravityla:
CI is actually passing on this branch but the status shows as failed.  It's a bug.  Exaggerated _eventual_ consistency

@ebkalderon I've hijacked this PR to instead pass `rust-src` into the results of `makePackageSet` and the `mkShell` so that there's a faster path to `rust-src` inclusion per-project for powering rust-analyzer having to make another channel (and keep the versions the same) and get it from there.

The conclusion I came to is that while sharing filenames in outputs is bad for constructing profiles for NixOS and home manager etc, the change needs to be made only when a package is bound for such a profile.  Removing `cargo-info` actually breaks builds.  The only way to detect that a user even wants to include the file in such a profile is up to that profile generation.  I have a workaround in place and will probably communicate the issue upstream if profile generation is not destined to have this issue.

This PR should still be safe because removing _empty_ `.link-flags` is supported in the various bash helpers, which check for existence or only use the file after touching the path.

trha:
I wonder if this is necessary, given that you can fix the home manager issue by extracting just the binaries:
```nix
binOnly = drv: runCommand "${drv.name}-bin" "mkdir -p $out && cp -r ${drv}/bin $out"
```

antigravityla:
> I wonder if this is necessary, given that you can fix the home manager issue by extracting just the executables:
> 
> ```nix
> binOnly = drv: runCommand "${drv.name}-bin" {} "mkdir -p $out && cp -r ${drv}/bin $out"
> ```

```nix
  package = (rustPkgs.workspace.rust-analyzer {}).overrideAttrs (
    oldAttrs: {
      installPhase = oldAttrs.installPhase + ''
      rm $out/.cargo-info
      rm $out/.link-flags || true
      '';
    }
  );
```
This was the hack I landed on first.

> A proper fix though, would allow crates to have multiple outputs, so you can refer to the executables with `<crate-name>.bin`.

I'll take a look at this pathway.  Sounds right.  Guess we need bin, lib, possibly more depending on context.

trha:
@antigravityla Let's remove the `rm` hack. We can merge the `rust-src` addition.

ebkalderon:
@antigravityla Is this PR being worked on still? Or has it been dropped in favored of adding dedicated `.bin` and `.lib` attributes to each crate derivation?

antigravityla:
> Or has it been dropped in favored of adding dedicated .bin and .lib attributes to each crate derivation?

Going this direction

PR #118, opened 2020-06-29
Use --message-format=json to install artifacts

trha:
### Changed

* Use `--message-format=json` to install artifacts on Rust >=1.41.0, falling back to the older method for compatibility with older versions of Rust that lack `out_dir` in the JSON output (https://github.com/rust-lang/cargo/pull/7622).
* Build examples in CI using the `HEAD` version of the overlay instead of the pinned versions, to actually check whether the current PR broke something.
* Address some `shellcheck` lints and clean up formatting of `overlay/utils.sh`.
* Lightly reorganize expression in `overlay/mkcrate.nix`.

This should hopefully fix the observed issues with recent versions (1.4X) of Rust placing test binaries in `target/deps` instead of `target`, thereby breaking the `install_crate` Bash function and the `cargo2nix` test runner.

@edude03 and @antigravityla, please confirm if `rustBuilder.runTests` now runs properly for you in CI with these changes!

### DIFF
--- a/overlay/mkcrate.nix
+++ b/overlay/mkcrate.nix
@@ -148,6 +148,12 @@ let
       EOF
     '';
 
+    configurePhase = ''
+      runHook preConfigure
+      runHook configureCargo
+      runHook postConfigure
+    '';
+
     manifestPatch = toJSON {
       features = genAttrs features (_: [ ]);
       profile.${ decideProfile compileMode release } = profile;
@@ -169,13 +175,6 @@ let
               } + $manifestPatch" \
         | remarshal -if json -of toml > Cargo.toml
     '';
-
-    configurePhase =
-      ''
-        runHook preConfigure
-        runHook configureCargo
-        runHook postConfigure
-      '';
 
     runCargo = ''
       (
--- a/overlay/mkcrate.nix
+++ b/overlay/mkcrate.nix
@@ -176,22 +176,6 @@ let
         | remarshal -if json -of toml > Cargo.toml
     '';
 
-    runCargo = ''
-      (
-        set -euo pipefail
-        if (( NIX_DEBUG >= 1 )); then
-          set -x
-        fi
-        env \
-          "CC_${stdenv.buildPlatform.config}"="${ccForBuild}" \
-          "CXX_${stdenv.buildPlatform.config}"="${cxxForBuild}" \
-          "CC_${host-triple}"="${ccForHost}" \
-          "CXX_${host-triple}"="${cxxForHost}" \
-          "''${depKeys[@]}" \
-          ${buildCmd}
-      )
-    '';
-
     setBuildEnv = ''
       MINOR_RUSTC_VERSION="$(${rustc}/bin/rustc --version | cut -d . -f 2)"
 
@@ -233,6 +217,22 @@ let
           echo $key
         done
       fi
+    '';
+
+    runCargo = ''
+      (
+        set -euo pipefail
+        if (( NIX_DEBUG >= 1 )); then
+          set -x
+        fi
+        env \
+          "CC_${stdenv.buildPlatform.config}"="${ccForBuild}" \
+          "CXX_${stdenv.buildPlatform.config}"="${cxxForBuild}" \
+          "CC_${host-triple}"="${ccForHost}" \
+          "CXX_${host-triple}"="${cxxForHost}" \
+          "''${depKeys[@]}" \
+          ${buildCmd}
+      )
     '';
 
     buildPhase = ''
--- a/overlay/mkcrate.nix
+++ b/overlay/mkcrate.nix
@@ -95,9 +95,8 @@ let
     runtimeDependencies buildtimeDependencies;
 
   drvAttrs = {
-    inherit NIX_DEBUG;
+    inherit src version meta NIX_DEBUG;
     name = "crate-${name}-${version}${optionalString (compileMode != "build") "-${compileMode}"}";
-    inherit src version meta;
     buildInputs = runtimeDependencies;
     propagatedBuildInputs = concatMap (drv: drv.propagatedBuildInputs) runtimeDependencies;
     nativeBuildInputs = [ cargo ] ++ buildtimeDependencies;
--- a/overlay/mkcrate.nix
+++ b/overlay/mkcrate.nix
@@ -80,7 +80,8 @@ let
     in
       ''
         cargo build $CARGO_VERBOSE ${optionalString release "--release"} --target ${host-triple} ${buildMode} \
-          ${featuresArg} ${optionalString (!hasDefaultFeature) "--no-default-features"}
+          ${featuresArg} ${optionalString (!hasDefaultFeature) "--no-default-features"} \
+          --message-format=json | tee .cargo-build-output
       '';
 
   inherit
@@ -193,10 +194,6 @@ let
     '';
 
     setBuildEnv = ''
-      isProcMacro="$( \
-        remarshal -if toml -of json Cargo.original.toml \
-        | jq -r 'if .lib."proc-macro" or .lib."proc_macro" then "1" else "" end' \
-      )"
       crateName="$(
         remarshal -if toml -of json Cargo.original.toml \
         | jq -r 'if .lib."name" then .lib."name" else "${replaceChars ["-"] ["_"] name}" end' \
@@ -236,7 +233,7 @@ let
     installPhase = ''
       mkdir -p $out/lib
       cargo_links="$(remarshal -if toml -of json Cargo.original.toml | jq -r '.package.links | select(. != null)')"
-      install_crate ${host-triple} ${if release then "release" else "debug"}
+      install_crate2 ${host-triple}
     '';
   };
 in
--- a/overlay/mkcrate.nix
+++ b/overlay/mkcrate.nix
@@ -194,14 +194,26 @@ let
     '';
 
     setBuildEnv = ''
+      MINOR_RUSTC_VERSION="$(${rustc}/bin/rustc --version | cut -d . -f 2)"
+
+      if (( MINOR_RUSTC_VERSION < 41 )); then
+        isProcMacro="$(
+          remarshal -if toml -of json Cargo.original.toml \
+          | jq -r 'if .lib."proc-macro" or .lib."proc_macro" then "1" else "" end' \
+        )"
+      fi
+
       crateName="$(
         remarshal -if toml -of json Cargo.original.toml \
         | jq -r 'if .lib."name" then .lib."name" else "${replaceChars ["-"] ["_"] name}" end' \
       )"
+
       . ${./utils.sh}
+
       export CARGO_VERBOSE=`cargoVerbosityLevel $NIX_DEBUG`
       export NIX_RUST_METADATA=`extractHash $out`
       export CARGO_HOME=`pwd`/.cargo
+
       mkdir -p deps build_deps
       linkFlags=(`makeExternCrateFlags $dependencies $devDependencies`)
       buildLinkFlags=(`makeExternCrateFlags $buildDependencies`)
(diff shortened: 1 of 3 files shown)


overlay/mkcrate.nix
```diff
@@ -231,7 +228,7 @@ let
     installPhase = ''
       mkdir -p $out/lib
       cargo_links="$(remarshal -if toml -of json Cargo.original.toml | jq -r '.package.links | select(. != null)')"
-      install_crate ${host-triple}
+      install_crate2 ${host-triple}
```
ebkalderon, line 231: Do you think we could somehow check the `rustc` or `cargo` versions and fall back to the old behavior on older versions of Rust? I see that `install_crate` has been kept around and not removed.

ebkalderon: Perhaps something like this?

```suggestion
      MINOR_RUSTC_VERSION="$(rustc --version | cut -d . -f 2)"
      if [[ "${MINOR_RUSTC_VERSION}" -ge 43 ]]; then
          install_crate2 ${host-triple}
      else
          install_crate ${host-triple}
      fi
```

overlay/mkcrate.nix
```diff
@@ -188,10 +189,6 @@ let
     '';
 
     setBuildEnv = ''
-      isProcMacro="$( \
-        remarshal -if toml -of json Cargo.original.toml \
-        | jq -r 'if .lib."proc-macro" or .lib."proc_macro" then "1" else "" end' \
-      )"
```
ebkalderon, line 194: If we choose to add a fallback method for earlier versions of Rust, as described above, we will need to restore these removed lines for that fallback case.

overlay/utils.sh
```diff
@@ -114,6 +115,60 @@ dumpDepInfo() {
...
+        local key="${BASH_REMATCH[1]}"
+        local val="${BASH_REMATCH[2]}"
+        printf 'DEP_%s_%s=%s\n' "$(upper "$cargo_links")" "$(upper "$key")" "$val" >> "$out/lib/.dep-keys"
+    done || :
+
+    if [[ -d "$out_dir" ]] && { grep "$out_dir" "$out/lib/.dep-keys" || grep "$out_dir" "$out/lib/.link-flags"; }; then
+        local installed_out_dir="$out/build_output" # TODO: Change me
+        sed -i "s#$out_dir#$installed_out_dir#g" "$out/lib/.dep-keys"
+        sed -i "s#$out_dir#$installed_out_dir#g" "$out/lib/.link-flags"
+        cp -r "$out_dir" "$installed_out_dir"
```
edude03, line 155: I noticed a bunch of log spam from I believe this line, where it seems to be trying to copy the `bin` folder from crates that don't have a build folder. I'll have to do a clean build to get the log output again, but see if you happen to notice that during your development as well

ebkalderon:
As discussed with @trha, I'm taking ownership of this PR for now.

overlay/utils.sh
ebkalderon: Interestingly, I noticed the log spam about the failed `bin` copy when building this very PR on my machine, @edude03. Take a look at this:

```
installing
cp: missing destination file operand after '/nix/store/a2azz41rpf73mdwfzv4chiffhdh1a9jz-crate-memchr-2.2.1/bin'
Try 'cp --help' for more information.
post-installation fixup
strip is /nix/store/69in1slwg74lqz3lzz15jbrck2rik21v-cctools-binutils-darwin/bin/strip
stripping (with command strip and flags -S) in /nix/store/a2azz41rpf73mdwfzv4chiffhdh1a9jz-crate-memchr-2.2.1/bin
patching script interpreter paths in /nix/store/a2azz41rpf73mdwfzv4chiffhdh1a9jz-crate-memchr-2.2.1
cp: missing destination file operand after '/nix/store/9ri7f4ygdih03npa75czwjx9qi0pcq5a-crate-opener-0.4.1/bin'
Try 'cp --help' for more information.
grep: /../output: No such file or directory
post-installation fixup
strip is /nix/store/69in1slwg74lqz3lzz15jbrck2rik21v-cctools-binutils-darwin/bin/strip
stripping (with command strip and flags -S) in /nix/store/9ri7f4ygdih03npa75czwjx9qi0pcq5a-crate-opener-0.4.1/bin
patching script interpreter paths in /nix/store/9ri7f4ygdih03npa75czwjx9qi0pcq5a-crate-opener-0.4.1
cp: missing destination file operand after '/nix/store/i5p1szlajv474rclnvyyvdhsggd3mh33-crate-pathdiff-0.1.0/bin'
Try 'cp --help' for more information.
grep: /../output: No such file or directory
post-installation fixup
strip is /nix/store/69in1slwg74lqz3lzz15jbrck2rik21v-cctools-binutils-darwin/bin/strip
stripping (with command strip and flags -S) in /nix/store/i5p1szlajv474rclnvyyvdhsggd3mh33-crate-pathdiff-0.1.0/bin
patching script interpreter paths in /nix/store/i5p1szlajv474rclnvyyvdhsggd3mh33-crate-pathdiff-0.1.0
cp: missing destination file operand after '/nix/store/s47ig50zjv67bynv9qhb0hprgqfy72kz-crate-percent-encoding-2.1.0/bin'
Try 'cp --help' for more information.
```

overlay/utils.sh
```diff
@@ -114,6 +115,60 @@ dumpDepInfo() {
...
+    # shellcheck disable=SC2046
+    cp $(jq -rR 'fromjson?
+        | select(.reason == "compiler-artifact")
+        | .filenames
+        | map(select(test("\\.(a|rlib|so|dylib)$")))
+        | .[]' < .cargo-build-output) "$out/lib" || :
+
+    mkdir -p "$out/bin"
+    # shellcheck disable=SC2046
+    cp $(jq -rR 'fromjson?
```
trha, line 129: The errors actually originate from this line if there are no binaries to copy. We can safely silent them by redirecting stderr to /dev/null

overlay/utils.sh
```diff
@@ -114,6 +115,60 @@ dumpDepInfo() {
     done
 }
 
+install_crate2() {
+    mkdir -p "$out/lib"
+    # shellcheck disable=SC2046
+    cp $(jq -rR 'fromjson?
```
trha, line 121: Same here for lib.

ebkalderon:
@trha What about the `grep` errors coming from `$out_dir` being undefined? Can we not guard certain parts by whether `out_dir` is set or not?

overlay/utils.sh
ebkalderon: We pre-emptively create a `bin` and `lib` directory regardless of whether `bin` or `lib` targets actually exist. I don't think we should be doing that, no?

ebkalderon: Or are these empty directories supposed to be present regardless, by Nixpkgs convention?

ebkalderon: @trha Also, what about the `grep` errors coming from `$out_dir` being undefined? Can we not guard certain parts of `install_crate2` by whether `out_dir` is set or not?

overlay/mkcrate.nix
ebkalderon: Done in c5f4107.

overlay/mkcrate.nix
ebkalderon: Done in c5f4107, albeit in a slightly different form.

overlay/utils.sh
trha: We can check if they're empty and remove them later. It's more straightforward that way.

ebkalderon: But `lib` isn't ever empty because `.dep-keys` is created there beforehand. And we'd have to go in and ignore errors for many different pieces of the function. I think it would actually be much easier to gate the creation of these directories based on whether `$out_dir` is set or not.

ebkalderon: Okay, I think I get it now, after some review. I think we can do both approaches. I believe that the two blocks you highlighted can be safely have `2> /dev/null` with no consequence. Likewise, the `grep` and `installed_out_dir` sections can be gated by whether `$out_dir` was set successfully or not.

overlay/utils.sh
ebkalderon: Fixed in a095422 and 0e258d2.

overlay/utils.sh
ebkalderon: Done in a095422 and 0e258d2.

ebkalderon:
Now ready for review! I can't mark @trha as a reviewer since he originally opened it, but would you mind reviewing it as well?

overlay/utils.sh
```diff
@@ -1,206 +1,283 @@
...
-    echo ${name##*.}
+  local name=$(basename "$1")
+  echo "${name##*.}"
 }
+
 extractHash() {
-    local name=`basename $1`
-    echo ${name%%-*}
+  local name=$(basename "$1")
+  echo "${name%%-*}"
```
edude03, line 10: Super minor nitpick, but I wonder if leaving basename is better? Definitely bikeshedding but I had to go off and read [this](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02) and [this](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02) to make sure that that's correct. That or leave a comment for people unfamilar with bash pattern expansion syntax?

ebkalderon: We can't use `basename` alone because that would produce `crate-foo-X.Y.Z-################` and we need only `################`. Sure, we could add a comment here, but the purpose of this Bash pattern is ultimately made redundant by the function name itself, `extractHash`.

edude03:
Ah yeah I see what you mean! OK, leave it as is then

overlay/mkcrate.nix
```diff
@@ -236,7 +243,11 @@ let
     installPhase = ''
       mkdir -p $out/lib
       cargo_links="$(remarshal -if toml -of json Cargo.original.toml | jq -r '.package.links | select(. != null)')"
-      install_crate ${host-triple} ${if release then "release" else "debug"}
+      if (( MINOR_RUSTC_VERSION < 41 )); then
+        install_crate ${host-triple} ${if release then "release" else "debug"}
+      else
+        install_crate2 ${host-triple}
```
trha, line 249: `install_crate2` doesn't need the `${host-triple}` argument so we can safely remove it.

overlay/utils.sh
```diff
@@ -1,206 +1,283 @@
...
-                if [ -e "$val" ]; then
-                    local dep_file_target=$dep_files/DEP_$(upper $cargo_links)_$(upper $key)
-                    cp -r "$val" $dep_file_target
-                    val=$dep_file_target
-                fi
-                printf 'DEP_%s_%s=%s\n' $(upper $cargo_links) $(upper $key) "$val" >>$dep_keys
-        esac
-    done
+    if [[ -d "$out_dir" ]] && { grep "$out_dir" "$out/lib/.dep-keys" || grep "$out_dir" "$out/lib/.link-flags"; }; then
+      local installed_out_dir="$out/build_output" # TODO: Change me
```
trha, line 169: I was not sure what this directory should be named. If this name looks good to you, we can remove the TODO.

ebkalderon: I'm honestly not sure what to make of this directory. What exactly is `$out/build_output` for? Why does it contain the contents of `bin` and `lib`?

ebkalderon: Would you mind elaborating what the purpose of this was, @trha?

ebkalderon: For future readers: it holds the contents of `$OUT_DIR` after a Cargo build script was run. These are often the result of build-time codegen or compiled non-Rust native libraries to be linked into the final crate.

Maybe we could rename it to `out_dir` in a subsequent PR to better clarify its purpose?

overlay/mkcrate.nix
ebkalderon: Sure! I'll do that in a follow-up PR.

overlay/utils.sh
ebkalderon: @trha had an even better suggestion, in my mind, which is `build_script_output`. This should make it even more clear what this directory contains. Considering that it's an implementation detail of `cargo2nix`, it shouldn't matter too much what it's called, but that sounds clear enough to me.

Let's merge these changes in as-is, and I'll open a follow-up PR in the meantime.

PR #119, opened 2020-07-06
Support for specifying target-feature and target-cpu for a package set

philipstears:
Howdy!

We're using cargo2nix, but ran into a snag. Our particular project makes use of AES via the "aes" crate. That particular crate uses conditions based on target-feature to decide whether to pull in a hardware-optimized crate, or use a software implementation instead:

```
[target.'cfg(not(all(target_feature="aes", target_feature = "sse2", any(target_arch = "x86_64", target_arch = "x86"))))'.dependencies]
aes-soft = { version = "0.4", path = "aes-soft" }
```

Today, cargo2nix doesn't support that.

I've created this PR to invite discussion about how that should be supported. For now, I've implemented a simple solution, but it isn't without drawbacks. The main one being that it is a breaking change currently, if this PR were adopted, folks using it would need to regenerate their Cargo.nix files to pick up the new `hostPlatformCpu` and `hostPlatformFeatures` variables.

I'm also not sure about the names of those variables, I picked them to split the difference between nix's `hostPlatform` concept, and rust's `target-cpu` and `target-feature` naming.

An example of using these new variables is as follows:

```
  rustPackages = super.rustBuilder.makePackageSet' {
    rustChannel = "stable";

    localPatterns = [
        # Standard included paths
        ''^(src|tests)(/.*)?''
        ''[^/]*\.(rs|toml)$''

        # Also include release-data folder
        ''^release-data(/.*)?''
    ];

    hostPlatformCpu = "sandybridge";

    hostPlatformFeatures = [
      "aes"
      "sse2"
      "ssse3"
      "sse4.1"
    ];
  };

```

### DIFF
--- a/overlay/make-package-set/full.nix
+++ b/overlay/make-package-set/full.nix
@@ -17,6 +17,8 @@
   fetchCrateAlternativeRegistry ? _: throw "fetchCrateAlternativeRegistry is required, but not specified in makePackageSet",
   release ? null,
   rootFeatures ? null,
+  hostPlatformCpu ? null,
+  hostPlatformFeatures ? [],
 }:
 lib.fix' (self:
   let
@@ -52,6 +54,8 @@ lib.fix' (self:
       };
       ${ if release == null then null else "release" } = release;
       ${ if rootFeatures == null then null else "rootFeatures" } = rootFeatures;
+      ${ if hostPlatformCpu == null then null else "hostPlatformCpu" } = hostPlatformCpu;
+      ${ if hostPlatformFeatures == null then null else "hostPlatformFeatures" } = hostPlatformFeatures;
     });
 
   in packageFunWith { mkRustCrate = mkRustCrate'; buildRustPackages = buildRustPackages'; } // {
--- a/Cargo.nix
+++ b/Cargo.nix
@@ -9,6 +9,8 @@
   rustPackages,
   buildRustPackages,
   hostPlatform,
+  hostPlatformCpu ? null,
+  hostPlatformFeatures ? [],
   mkRustCrate,
   rustLib,
   lib,
@@ -35,7 +37,7 @@ in
     registry = "registry+https://github.com/rust-lang/crates.io-index";
     src = fetchCratesIo { inherit name version; sha256 = "5d2e7343e7fc9de883d1b0341e0b13970f764c14101234857d2ddafa1cb1cac2"; };
   });
-  
+
   "registry+https://github.com/rust-lang/crates.io-index".aho-corasick."0.7.6" = overridableMkRustCrate (profileName: rec {
     name = "aho-corasick";
     version = "0.7.6";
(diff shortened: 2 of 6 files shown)


ebkalderon:
@dingxiangfei2009 Seems like we should. Could we update [these comments](https://github.com/tenx-tech/cargo2nix/blob/master/default.nix#L26-L45) in our `default.nix` to clarify this point? I might open a PR later and add a new tutorial to the `examples` directory demonstrating this point using the same `aes` crate as well.

ebkalderon:
@philipstears Would you mind briefly documenting the new optional `hostPlatformCpu` and `hostPlatformFeatures` arguments [in this comment block](https://github.com/tenx-tech/cargo2nix/blob/master/default.nix#L26-L45), please? Just a sentence or two should suffice for now. We can always expand on this documentation in subsequent PRs with a proper tutorial on using target-specific features with `cargo2nix`.

philipstears:
Hi @ebkalderon - sure thing, I've added a few lines explaining the options.

philipstears:
> I don't think it's possible to avoid the breakage you mentioned, sadly, since `stdenv` in Nixpkgs doesn't seem to have an equivalent to Rust's `target-cpu` and `target-features`, and the inputs will likely need to be provided by the user at the top-level. 

This was the conclusion I came to as well. The only other option I could see was to maybe augment `hostPlatform` in some way, but that just felt wrong to me because it's a core nixpkgs facility.

overlay/make-package-set/full.nix
```diff
@@ -17,6 +17,8 @@
   fetchCrateAlternativeRegistry ? _: throw "fetchCrateAlternativeRegistry is required, but not specified in makePackageSet",
   release ? null,
   rootFeatures ? null,
+  hostPlatformCpu ? null,
```
trha, line 20: We might need to provide a different set of features and CPU for the build platform in `make-package-set/simplified.nix`, but I doubt that there's any crate out there makes use of it, so I guess this should be fine.

ebkalderon: Would you mind opening a GitHub issue for this, @trha? We should revisit it for completeness' sake.

ebkalderon:
I think this should be good to merge now. We'll have to extend it to include separate CPU settings for the build platform, but the basic concept here is done. Thanks again for working on this, @philipstears! This is a very useful addition.

philipstears:
Thanks @ebkalderon and @trha!

When you mention CPU settings for the build platform, what do you mean specifically? it's possible to tell rust to use the build CPU as the host CPU by setting `target-cpu` to `"native"`, do you mean something different than that?

ebkalderon:
I'm assuming`target-cpu=native` was what @trha meant as well, @philipstears. That's why I would like a new issue to be opened to further elaborate on the matter. 😄

PR #80, opened 2020-02-17
Convert Cargo.nix codegen to Tera

ebkalderon:
### Changed

* Convert `Cargo.nix` templating logic to use [Tera](https://crates.io/crates/tera), replacing `fmt_ext`.

### Removed

* Remove unused `resolve.rs` module.

---

This should greatly simplify our Rust rendering logic, with little performance impact on release builds but a noticeable slowdown in debug builds.

To test the integrity of the changes in this commit, a regression test was created by wrapping `diesel`, a complex and feature-heavy crate workspace, with `cargo2nix` and generating a `Cargo.nix` with the latest `master` (0.8.1-unreleased) and a `Cargo.nix2` generated with
this commit.

Running `diff Cargo.nix Cargo.nix2` confirms that the only changes between the two versions are:

1. The addition of a newline between the top comment and the Nix expression.
2. The removal of an unnecessary trailing space from each line of `rootFeatures`.
3. The unquoting of `unknown` in members of `workspace`.
4. The removal of an unnecessary trailing newline from the bottom of the file.

Since no breaking changes were introduced, this refactor can be considered a non-breaking change, worthy of a patch version bump.

Closes #71.

PR #121, opened 2020-07-07
Escape embedded TOML in profilesByName properly

### LINKED ISSUE #120: profilesByName emitting doesn't escape strings
Howdy.

I tried adding:

```
[profile.release]
lto = "fat"
```

To my Cargo.toml, but when cargo2nix generates Cargo.nix, it emits:

```
   profilesByName = {
     release = builtins.fromTOML "lto = "fat"
 ";
   };
```

I'm unclear about the redundant new-line, but the main thing here is that the quotes inside of the TOML need escaping somehow, presumably before being passed to the template.

As a workaround, I'm able to use `lto = true` instead, but there are other potentially string-valued parameters that folks might want to change, for example the `panic` parameter.

ebkalderon:
### Fixed

* Escape TOML output in `profilesByName` using a combination of `str::{escape_debug,replace}`.

The Tera templating code for `profilesByName` hadn't been rigorously tested, and as such, it didn't perform proper escaping of sensitive characters (e.g. `${`, `"`, and so on) before embedding the unsanitized TOML into a double-quoted Nix string in the `Cargo.nix`. This string might also contain newlines or other control characters which would need to be escaped as well.

This commit utilizes `String::escape_debug()` to perform the bulk of the escaping work, sanitizing common control characters such as `\n`, `\r`, and `\t`, and then we also replace all instances of `${` with `\${` to avoid accidental string interpolation of variables.

This bug was first introduced in #80, when we first ported `cargo2nix` to use Tera for templating.

Fixes #120. CC @philipstears

### DIFF
--- a/src/template.rs
+++ b/src/template.rs
@@ -31,6 +31,7 @@ impl BuildPlan {
         let profiles = profiles
             .into_iter()
             .map(|(name, profile)| (name, toml::to_string(&profile).unwrap()))
+            .map(|(name, toml)| (name, toml.replace("${", "\\${").escape_debug().to_string()))
             .collect();
 
         let workspace_members = root_pkgs


ebkalderon:
An alternative approach to this problem could be to replace the TOML with a multi-line `''` string in [`templates/Cargo.nix.tera:22`](https://github.com/tenx-tech/cargo2nix/blob/b09e88570f2eaa729b2ae8b47a6a82b5fcb28de2/templates/Cargo.nix.tera#L22), which could reduce the amount of codegen-time escaping needed. I'm looking into this.

ebkalderon:
Just had to rebase this branch after merging #119, and I needed to regenerate the `Cargo.nix` expressions and examples' version pins because they seem to have been accidentally destroyed mid-rebase while addressing changes on that PR. Would you mind reviewing again, @trha?

philipstears:
I just had a quick test of this against my original scenario, which was the following in my `Cargo.toml`

```
[profile.release]
lto = "fat"
```

And it worked perfectly. Thanks.

ebkalderon:
Sweet! Glad it fixed the problem, @philipstears.

PR #124, opened 2020-07-13
Remove unused sha256 argument for Git dependencies

### LINKED ISSUE #102: Consider using builtins.fetchGit over nix-prefetch-git
We currently make use of `fetchgitPrivate` inside the overlay for fetching Git dependencies, but [`fetchgitPrivate` has been removed](https://github.com/NixOS/nixpkgs/commit/e578b8499f9e49441d9761875659381f95975fee) from newer versions of Nixpkgs.

Since `builtins.fetchGit` exists since Nix 2.0 and is capable of fetching Git repos at evaluation time without needing a `sha256` value, and it even works with private Git repos in much the same manner as `fetchgitPrivate` once did, there isn't much reason to stick with `nix-prefetch-git` anymore ([source](https://stackoverflow.com/a/49004755)). We should probably drop our dependency on `nix-prefetch-git` and `serde_json` and use `builtins.fetchGit` in the overlay instead.

Note that this would be a breaking change to the overlay since as we would have to remove the `sha256` argument from `fetchCrateGit`. This would require us to release a new minor version.

ebkalderon:
### Changed

* Simplify `checksum` field of `ResolvedPackage` from `Option<Cow<'a, str>` to `Option<&'a str>`.

### Removed

* Remove unused `sha256` argument for Git dependencies.
* Remove `serde_json` dependency.
* Remove private `prefetch_git` function.
* Remove `nix-prefetch-git` wrapper from `default.nix`.

This argument has been rendered unused since the transition to `builtins.fetchGit` for all Git dependencies was made (#113). Since a minor release is coming soon, due to #119, this breaking change should now be safe to roll out.

Closes #102.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -189,7 +189,6 @@ dependencies = [
  "pathdiff 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "semver 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.104 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde_json 1.0.41 (registry+https://github.com/rust-lang/crates.io-index)",
  "tempfile 3.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "tera 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "toml 0.5.3 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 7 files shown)


ebkalderon:
@trha Sure! I'll do that next. I also noticed that I can turn `checksum` into a `&'a str` instead of a `Cow<'a, str>`, now that the `Owned` case has been eliminated, so I'll do that as well.

ebkalderon:
Done! Would you mind re-reviewing, please, @trha?

ebkalderon:
Okay, merging this in. I imagine this should be okay, since we've been awaiting this transition for a long time now.

PR #125, opened 2020-07-13
Fix install_crate behavior for debug builds

ebkalderon:
### Fixed

* Switch to the `target/${triple}/release` or `target/${triple}/debug` directory when installing a crate, now properly respecting the `release` boolean flag.

Note that this fix should be rendered unnecessary by #118, but we still need to support older versions of Rust <1.41.0.

### DIFF
--- a/overlay/mkcrate.nix
+++ b/overlay/mkcrate.nix
@@ -236,7 +236,7 @@ let
     installPhase = ''
       mkdir -p $out/lib
       cargo_links="$(remarshal -if toml -of json Cargo.original.toml | jq -r '.package.links | select(. != null)')"
-      install_crate ${host-triple}
+      install_crate ${host-triple} ${if release then "release" else "debug"}
     '';
   };
 in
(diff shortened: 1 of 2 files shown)


ebkalderon:
Rebased after merging #127.

ebkalderon:
Rebased after merging #128.

antigravityla:
> Note that this fix should be rendered unnecessary by #118, but we still need to support older versions of Rust <1.41.0.

I assume you mean that this change should be gated behind an if clause that will only get finalized in #118.  Let's park this PR for now?

ebkalderon:
@antigravityla No, it should be just as valuable to merge into `master` with or without #118. However, it will only be relevant for Rust versions <1.41.0 once #118 is merged, as #118 happens to fix the same issue for Rust >=1.41.0 for free.

ebkalderon:
Rebased after merging #126.

ebkalderon:
Good idea, @trha! I wonder if it needs a dedicated example project for this, or if we can build the existing projects in both debug and release mode in CI? I might open a PR for this later, and we can discuss there, if you'd like.

PR #126, opened 2020-07-13
Validate shell.nix and mkcrate-nobuild.nix in CI

ebkalderon:
### Changed

* Invoke `nix-shell` in CI to verify the validity of `shell.nix` and `overlay/mkcrate-nobuild.nix`.

We've been bitten by this before (#123), so I think this change would come in handy.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,6 +19,7 @@ jobs:
             cachix push cargo2nix -w &
             nix-build -A ci --show-trace
             nix-build -A examples -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/47b551c6a854a049045455f8ab1b8750e7b00625.tar.gz --show-trace
+            nix-shell --show-trace --run exit
           no_output_timeout: 5h
       - save_cache:
           key: nix-store


ebkalderon:
Rebased after merging #127.

ebkalderon:
Rebased after merging #128.

.circleci/config.yml
```diff
@@ -19,6 +19,7 @@ jobs:
             cachix push cargo2nix -w &
             nix-build -A ci --show-trace
             nix-build -A examples -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/47b551c6a854a049045455f8ab1b8750e7b00625.tar.gz --show-trace
+            nix-shell --no-build-output --show-trace --run exit
```
antigravityla, line 22: I feel like --no-build-output may have slipped in somewhere... it doesn't seem to be appropriate for CI.

>        --no-build-output / -Q
           By default, output written by builders to standard output and
           standard error is echoed to the Nix command's standard error. This
           option suppresses this behaviour. Note that the builder's standard
           output and error are always written to a log file in
           prefix/nix/var/log/nix.

At your discretion.  Pre-approved either way.

ebkalderon: @antigravityla I copy-pasted it from one of our other repositories which makes use of `nix-shell` validation in CI. 😬 Looking at the docs, the flag doesn't really seem necessary here. I can remove it.

ebkalderon: Thanks for the heads-up!

ebkalderon: Done and force-pushed! Would you mind re-reviewing again, @antigravityla?

ebkalderon:
Given that it's been a few days with no objections, I'm merging this PR in.

PR #127, opened 2020-07-13
Fix indentation in overlay/mkcrate.nix

ebkalderon:
As the title says. This has apparently been around for a long time.


ebkalderon:
Thanks, @antigravityla! 🤗

PR #128, opened 2020-07-13
Update outdated Git tag in example project READMEs

ebkalderon:
### Fixed

* Update READMEs of example projects to reference `v0.8.3` instead of outdated versions. This text should have been updated back when 0.8.3 was released, but it seems it slipped through.


ebkalderon:
Sorry for the second rubber-stamping, guys. This should be another very small and uncontroversial change, so I appreciate the quick review, @antigravityla.

PR #135, opened 2020-07-19
Add tutorial for static resources example project

ebkalderon:
### Added

* Add tutorial for static resources example project.

### Changed

* Update end of `README.md` for `2-bigger-project` for proper continuity.
* Rename `3-project-with-resources` to `3-static-resources`.
* Fix indentation and simplify `localPatterns` expression in `3-static-resources`.
* Remove auto-generated comments from `Cargo.toml` files in examples.
* Preemptively reference `v0.9.0` tag in tutorials.
* Update example projects' `cargo2nix` pins to a newer version compatible with #119.

Closes #106.

PR #136, opened 2020-07-23
Follow-up fixes for static resources project tutorial

ebkalderon:
### Fixed

* Correct typo "eand" -> "and".

### Removed

* Remove noisy `result` line from filtering table.

Follow-up to #135 which addresses feedback given by @antigravityla on that PR.

### DIFF
--- a/examples/3-static-resources/README.md
+++ b/examples/3-static-resources/README.md
@@ -120,7 +120,7 @@ fn main() {
 
 This build script will generate a Rust source file named `$OUT_DIR/templates.rs`
 which we can statically embed into our program with `include_str!()`. Let's open
-the `src/main.rs` file eand write some code that will do just that:
+the `src/main.rs` file and write some code that will do just that:
 
 ```rust
 //! A program which generates some text using an embedded template.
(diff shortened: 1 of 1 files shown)


examples/3-static-resources/README.md
```diff
@@ -11,28 +11,6 @@ to only a few specific files and directories at build time. Specifically:
...
-│   └── main.rs
-├── tests
-│   └── my_tests.rs
-├── build.rs
-├── default.nix
-├── Cargo.lock
-├── Cargo.nix
-├── Cargo.toml
-├── README.md
-└── result -> /nix/store/qjz1m7pmm23r7fhgbigb2x5pnwqp925j-crate-foo-project-0.1.0
```
antigravityla, line 29: The nit I left was only for the `result` line.  The tree itself is helpful

ebkalderon: I see; my mistake! I can restore the tree without the `result` entry, in that case. Would that work, @antigravityla?

ebkalderon: Okay, amended the last commit!

ebkalderon:
Anytime, @antigravityla! Thanks for the suggestions. Sorry I couldn't fit them into the other PR, but there were so many approvals already, I didn't want to invalidate them all for such minor changes. 😅

PR #138, opened 2020-07-27
Add rust src to package set

antigravityla:
This shortcut makes it easier to construct a shell with coordinated rust sources
for consumption in rust analyzer

The good part of https://github.com/tenx-tech/cargo2nix/pull/115/files

### DIFF
--- a/overlay/make-package-set/full.nix
+++ b/overlay/make-package-set/full.nix
@@ -11,6 +11,7 @@
   packageFun,
   cargo,
   rustc,
+  rust-src,
   buildRustPackages ? null,
   localPatterns ? [ ''^(src|tests)(/.*)?'' ''[^/]*\.(rs|toml)$'' ],
   packageOverrides ? rustBuilder.overrides.all,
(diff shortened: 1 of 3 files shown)


antigravityla:
Lost at least one build to Ci.. local succeeds.  Cautious

overlay/make-package-set/full.nix
```diff
@@ -11,6 +11,7 @@
   packageFun,
   cargo,
   rustc,
+  rust-src,
```
trha, line 14: Let's just get rid of these and take `rustChannel` for simplicity.

antigravityla: Just to simplify arguments?

antigravityla:
> Is this a good idea? Do you know if this would lead to each crate derivation pulling in rust-src when copied between Nix stores when it's not strictly speaking needed, depending on whether we're cross-compiling or not?

I'm leaning on lazy evaluation AFAIK, so only if someone is consuming the attribute, but please elaborate if I've misinterpreted the behavior in the case that the user never writes `rust-src` anywhere

psionic-k:
@antigravityla  closing in favor of upcoming rust-analyzer packaging & proper shell support

@ebkalderon to address your earlier concern I'll take a look at the closures for bin targets for example.

Upcoming PR will encompass concerns from #151 and #152, #145, this PR, and a few outstanding issues that make it difficult to employ cargo2nix as a solution for independent nix packaging of rust projects

Issue #145, opened 2020-10-08
Rust 1.47 - Cannot parse v2 Lockfile Format


cryptoquick:
So, the latest Rust version got pushed to stable last night, which includes a new lockfile format, and it breaks all our builds. We were able to fix it by replacing `stable` with `1.46.0`, but that's not ideal.

The error we got was:
`error: failed to parse lock file at: /build/bitflags-1.2.1/Cargo.lock`

It wasn't just bitflags, there were several packages that broke, but that's just an example.

ebkalderon:
Thanks for reporting this, @cryptoquick! No worries, I can regenerate the lockfile with the latest stable and make it part of the next release.

ebkalderon:
@psionic-k I think this may increase the urgency of resolving #110. In theory, we could simply regenerate the `Cargo.lock` under a newer version of `cargo` and push it to `master`, but then `cargo2nix` would not be able to bootstrap itself because it links against an older version of `cargo`. :disappointed: PR #137 was supposed to resolve this, but it has gone stale due to some strange compilation issues on Darwin.

ebkalderon:
I might give it another go today, seeing as I finally have time!

psionic-k:
You will likely get to the mac issues first.  I need to provision my Apple machine before I can jump in that race.

ebkalderon:
Sounds good! I've just switched over from my Linux machine to my Mac to give it a try.

PR #137, opened 2020-07-27
Update cargo 0.41 -> 0.46

antigravityla:
Cherry picked everything.  Going to run back to the original PR and try to address outstanding comments

https://github.com/tenx-tech/cargo2nix/pull/111

@macs-overdrv

Closes #110

Issue #110, opened 2020-05-03
Update to latest cargo API encounters breaking changes


macs-overdrv:
Updating the cargo dependency is mostly straightforward except changes to one method cargo2nix uses, `cargo::ops::resolve_ws_with_opts` because of the new arguments it accepts

https://docs.rs/cargo/0.44.0/cargo/ops/fn.resolve_ws_with_opts.html
https://docs.rs/cargo/0.41.0/cargo/ops/fn.resolve_ws_with_opts.html

Most of the changes happened in this PR
https://github.com/rust-lang/cargo/pull/7820

Here's internal usages of those arguments:
https://github.com/rust-lang/cargo/search?q=HasDevUnits&unscoped_q=HasDevUnits
https://github.com/rust-lang/cargo/search?q=CompileKind&unscoped_q=CompileKind

I'm filing this issue because at a minimum, the API changes will block cargo2nix from using newer cargo features, meaning some projects with recent cargo features cannot be built with current cargo2nix

ebkalderon:
Thanks for reporting the issue, @macs-overdrv! ❤️

I remember seeing that PR around the the time that it had been merged, and I don't know if there is an easy solution to this. In particular, ignoring targets that aren't being built and not unifying features between normal, build, and dev dependencies break certain fundamental assumptions about how `cargo2nix` works and how we perform feature resolution.

CC @trha

macs-overdrv:
> In particular, ignoring targets that aren't being built and not unifying features between normal, build, and dev dependencies

The dev units issue appears to have a simple workaround.  [If dev units are always turned on](https://github.com/tenx-tech/cargo2nix/pull/111/commits/ac81eb2031ca8d715052cf67793a813309ea103d), the resulting `packageFun` might be a little bit heavy, but not broken.  Alternatively, since dev units are not expected to be useful for distribution, the role of cargo2nix, there are also strong arguments in favor of defaulting to excluding dev units.

The [signature in newer cargo](https://github.com/tenx-tech/cargo2nix/pull/111/files#diff-639fbc4ef05b315af92b4d836c31b023L299-R313) strongly suggests target-dependent features were added, which the newer cargo library can evaluate correctly.  However, the output of a package function for a _single_ target would only cover one possible result from cargo.

[...]

PR #146, opened 2020-11-21
Udev and dbus overrides

psionic-k:
Two simple overrides and one drive-by override fixup.

Use of `$RUSTC` will fail if we ever switch how it's provided to the build env.  Recommend a followup PR to always provide `rustc` on the `$PATH` for other creative build.rs scripts which attempt to invoke it directly

### DIFF
--- a/overlay/overrides.nix
+++ b/overlay/overrides.nix
@@ -67,6 +67,7 @@ in rec {
     fsevent-sys
     libgit2-sys
     libdbus-sys
+    libudev-sys
     openssl-sys
     pkg-config
     pq-sys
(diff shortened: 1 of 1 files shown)


ebkalderon:
@psionic-k I think the PR is failing due to #145. We should open a separate PR upgrading the `Cargo.lock` for `cargo2nix` and the included examples so it's compatible with Rust 1.47+.

ebkalderon:
Blocked on #148.

ebkalderon:
@psionic-k Okay, #148 is now merged! Try rebasing against `master` and give it another shot.

ebkalderon:
After this PR gets merged, let's try to get #137 over the finish line. Or perhaps restart work on it with a new branch and an even newer `cargo` version, if possible. If we can successfully complete that task, we will be able to resolve both #110 and #145 in one go, and we will be able to support the latest stable Rust again.

ebkalderon (dismissed):
Looks great! Just make sure to rebase against `master` so that it passes CI. Feel free to ping me if you need a re-review.

ebkalderon (approved):

PR #148, opened 2020-11-22
Temporarily pin examples to Rust 1.46.0

ebkalderon:
This should hopefully unblock CI for us until we can upgrade to a newer verison of Cargo which supports Lockfile format V2.

We can revert this once #145 is resolved.


psionic-k (approved):
:shipit:

PR #152, opened 2020-12-04
Configurable source

psionic-k:
This PR builds on #151 presumptive of merge

It's possible to do a test of providing a source by modifying the `default.nix` to point to some other source such as a git fetch of this branch or a copy of the cargo2nix source.  The Cargo.nix built with the result of such evaluation can then be used to self-host by regenerating Cargo.nix with `result/bin/cargo2nix -f` and removing the workspace files to test that only the external sources are being used to produce a new, working `Cargo.nix` and further another working `cargo2nix` in the result


psionic-k:
Closing in favor of #153

PR #165, opened 2020-12-23
switch from remarshal to yj

### LINKED ISSUE #149: "Cannot convert data to toml"
Hello! I'm using cargo2nix for the first time - and indeed compiling a Rust program for the first time - and it's starting off great but I'm hitting some errors like this:

```
unpacking sources
unpacking source archive /nix/store/43ksm0c42cmsy0am22vv2nm4rkqwbddx-byteorder-1.3.4.tar.gz
source root is byteorder-1.3.4
setting SOURCE_DATE_EPOCH to timestamp 1581081480 of file byteorder-1.3.4/Cargo.toml
patching sources
configuring
building
Error: Cannot convert data to TOML (Invalid type <class 'NoneType'>)
```

and I'm wondering if someone could suggest a way to troubleshoot this?

Background links:

- Rust repo that I'm trying to build: https://github.com/feenkcom/libskia
- [Gist](https://gist.github.com/099c3b68fa27271fa1e7f1a206df035e) containing `Cargo.lock`, `Cargo.nix`, `default.nix`.
- [Gisted Cargo.toml](https://gist.github.com/lukego/d3bf4990cb1571e8f9bc7b7779438d7a) referenced in the error above.
- Installed `cargo2nix` today according to the README on a recent NixOS master host.

Thanks for any pointers!

nightmared:
Fixes #149

### DIFF
--- a/overlay/mkcrate.nix
+++ b/overlay/mkcrate.nix
@@ -102,8 +102,8 @@ let
     nativeBuildInputs = [ cargo ] ++ buildtimeDependencies;
 
     depsBuildBuild =
-      let inherit (buildPackages.buildPackages) stdenv jq remarshal;
-      in [ stdenv.cc jq remarshal ];
+      let inherit (buildPackages.buildPackages) stdenv jq yj;
+      in [ stdenv.cc jq yj ];
 
     # Running the default `strip -S` command on Darwin corrupts the
     # .rlib files in "lib/".
@@ -167,7 +167,8 @@ let
         echo source = \"registry+''${registry}\" >> Cargo.lock
       fi
       mv Cargo.toml Cargo.original.toml
-      remarshal -if toml -of json Cargo.original.toml \
+
+      yj -t < Cargo.original.toml \
         | jq "{ package: .package
               , lib: .lib
               , bin: .bin
@@ -175,7 +176,8 @@ let
               , example: .example
               , bench: (if \"$registry\" == \"unknown\" then .bench else null end)
               } + $manifestPatch" \
-        | remarshal -if json -of toml > Cargo.toml
+        | jq '. | walk(if type == "object" then with_entries(select(.value != null)) else . end)' \
+        | yj -jt > Cargo.toml
     '';
 
     setBuildEnv = ''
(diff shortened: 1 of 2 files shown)


overlay/mkcrate.nix
```diff
@@ -167,29 +167,31 @@ let
...
+      yj -t < Cargo.original.toml \
         | jq "{ package: .package
               , lib: .lib
               , bin: .bin
               , test: .test
               , example: .example
               , bench: (if \"$registry\" == \"unknown\" then .bench else null end)
               } + $manifestPatch" \
-        | remarshal -if json -of toml > Cargo.toml
+        | jq '. | walk(if type == "object" then with_entries(select(.value != null)) else . end)' \
```
nightmared, line 179: This was a previous attempt at deleting JSON `null` values, but this gave no result. I will drop this.

nightmared:
The CI failure is weird, looks like this PR doesn't like the examples. I will try to see why.

nightmared:
Allright, this appears to come from https://github.com/sclevine/yj/issues/3, which is fixed in the latest version of yj, which is sadly not packaged yet in nixpkgs ;(

psionic-k:
So we're waiting on both solutions to potentially upstream.  Wish we had a `tq` and could avoid remarshalling.  Alternative ways to propagate the data between steps would also be welcome

nightmared:
It's even worse than I expected, because i just checked and yj 5.0.0 cannot parse cargo files:
`[target."cfg(target_os=\"windows\")".dependencies.winapi]` yields `Error parsing TOML: (41, 2): invalid table array key: invalid bare key character:`, so we are in a state where:
remarshal 0.14 and upper cannot parse the manifests
yj 4.0 can parse the file but conflates integers and floats
yj 5.0 and upper cannot parse the manifests either, but they properly differentiate integer and floats ;(

This is quite a mess ;( Indeed, it would be so great to have a `jq`-like tool for that !

Currently I see three ways out of this:
1 - Fix TOML Kit, and update remarshal to use the fix
2 - Fix https://github.com/pelletier/go-toml (or hopefully this was already implemented but yj relies on an older version ?)
3 - use both `yq` and `remarshal` to work around the loopholes in each (maybe this works, I don't know)
42 - Derive another way (a small rust crate using the same parser as cargo ?) to modify the toml without relying on external converters

psionic-k:
>  42 - Derive another way (a small rust crate using the same parser as cargo ?) to modify the toml without relying on external converters
 
 I'm in favor of this, likely sending the information down to dependencies via a build-specific dependency that will allow for more re-use of crates compiled with the exact same options.  There have been some crates released since Cargo2nix was originally developed that can query features etc and may allow elimination of the cargo dependency altogether (very attractive since cargo's API is mainly intended for cargo itself)

nightmared:
You're probably right, but I'm not exactly sure how that would compares to e.g. https://github.com/kolloch/crate2nix that already appear to do something similar.
Anyway, what I do know is that this PR won't get us there, since `yj` is not the solution to the issue at hand, so I'm closing this.
Good luck with that issue !

PR #171, opened 2021-01-09
strip nulls from the outputs of the first jq pipe

### LINKED ISSUE #149: "Cannot convert data to toml"

### LINKED ISSUE #167: User provided overrides broken
The documentation for `packageOverrides`:

```
  # - `packageOverrides` (optional):
  #     A function taking a package set and returning a list of overrides.
  #     Overrides are introduced to provide native inputs to build the crates generated in `Cargo.nix`.
  #     See `overlay/lib/overrides.nix` on how to create overrides and `overlay/overrides.nix` for a list of predefined overrides.
  #     Most of the time, you can just use `overrides.all`. You can hand-pick overrides later if your build becomes too slow.
```

To excite this problem, lets cherry pick from `rustBuilder.overrides`, which works just fine.  Then let's add a custom overlay that mimics the form in `overlay/overrides.nix`, but it breaks.

```nix
    inherit rustChannel;
    packageOverrides = pkgs: [
      pkgs.rustBuilder.overrides.capLints
      pkgs.rustBuilder.overrides.cc
      pkgs.rustBuilder.overrides.curl-sys
      pkgs.rustBuilder.overrides.libssh2-sys
      pkgs.rustBuilder.overrides.openssl-sys
      pkgs.rustBuilder.overrides.pkg-config
      pkgs.rustBuilder.overrides.rand
      pkgs.rustBuilder.overrides.rand_os
      pkgs.rustBuilder.overrides.ring
      # Previous overrides all work
      pkgs.rustBuilder.rustLib.makeOverride {
        name = "libdbus-sys";
        overrideAttrs = drv: {
          propagatedNativeBuildInputs = drv.propagatedNativeBuildInputs or [ ] ++ [
            pkgs.zlib
          ];
        };
      }
    ];
```

While this exact override could work when evaluated in `overlay/overrides.nix` it produces the following errors when provided like above:

```
nix-build -A package --show-trace
error: --- TypeError ----------------------------------------------------------------------------------------------- nix-build
at: (41:25) in file: /nix/store/sncgkl5q6k1skf9sawrlydhbyamzb6wv-source/overlay/lib/overrides.nix

    40|       leftOverriders = left args;
    41|       rightOverriders = right args;
      |                         ^
    42|     in

'makeOverride' at /nix/store/sncgkl5q6k1skf9sawrlydhbyamzb6wv-source/overlay/lib/overrides.nix:27:18 called with unexpected argument 'src'

```

Defining overrides in this position breaks.  Something about the combination logic has an issue or it's very unclear how to provide a custom override

### LINKED ISSUE #172: unsupported Git input attribute 'name'
Versions of nix have been spotted that don't support the (entirely optional) `name` argument to `fetchGit`
https://github.com/NixOS/nix/pull/3459/files#diff-c1137a9bb14eb1c2321ed480ee5843046d3a076fe03bfcdc5949f01f6cdcd7dbR57

You can update  or roll back the version of nix itself to work around this.  The offending lines don't exist in master.

Getting this into our issues for those searching.

psionic-k:
These were breaking the version of remarshal in use

Fixes #149 
Fixes #172
Fixes #167

Getting this into CI while I figure out what was broken.  All output from jq would have contained these nulls and it should have been breaking more

### DIFF
--- a/overlay/mkcrate.nix
+++ b/overlay/mkcrate.nix
@@ -174,7 +174,8 @@ let
               , test: .test
               , example: .example
               , bench: (if \"$registry\" == \"unknown\" then .bench else null end)
-              } + $manifestPatch" \
+              } | with_entries(select( .value != null ))
+              + $manifestPatch" \
         | remarshal -if json -of toml > Cargo.toml
     '';
 
--- a/overlay/mkcrate.nix
+++ b/overlay/mkcrate.nix
@@ -167,7 +167,16 @@ let
         echo source = \"registry+''${registry}\" >> Cargo.lock
       fi
       mv Cargo.toml Cargo.original.toml
-      # Remarshal version cannot handle escaped double quotes in table key paths
+      # Remarshal was failing on table names of the form:
+      # [key."cfg(foo = \"a\", bar = \"b\"))".path]
+      # The regex to find or deconstruct these strings must find, in order,
+      # these components: open bracket, open quote, open escaped quote, and
+      # their closing pairs.  Because each quoted path can contain multiple
+      # quote escape pairs, a loop is employed to match the first quote escape,
+      # which the sed will replace with a single quote equivalent until all
+      # escaped quote pairs are replaced.  The grep regex is identical to the
+      # sed regex but does not destructure the match into groups for
+      # restructuring in the replacement.
       while grep '\[[^"]*"[^\\"]*\\"[^\\"]*\\"[^"]*[^]]*\]' Cargo.original.toml; do
         sed -i -r 's/\[([^"]*)"([^\\"]*)\\"([^\\"]*)\\"([^"]*)"([^]]*)\]/[\1"\2'"'"'\3'"'"'\4"\5]/g' Cargo.original.toml
       done;
--- a/overlay/mkcrate.nix
+++ b/overlay/mkcrate.nix
@@ -180,6 +180,7 @@ let
               , bench: (if \"$registry\" == \"unknown\" then .bench else null end)
               } | with_entries(select( .value != null ))
               + $manifestPatch" \
+        | jq "del(.[][] | nulls)" \
         | remarshal -if json -of toml > Cargo.toml
     '';
 
--- a/overlay/mkcrate.nix
+++ b/overlay/mkcrate.nix
@@ -167,6 +167,10 @@ let
         echo source = \"registry+''${registry}\" >> Cargo.lock
       fi
       mv Cargo.toml Cargo.original.toml
+      # Remarshal version cannot handle escaped double quotes in table key paths
+      while grep '\[[^"]*"[^\\"]*\\"[^\\"]*\\"[^"]*[^]]*\]' Cargo.original.toml; do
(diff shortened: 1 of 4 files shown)


psionic-k:
Added some more munging after finding more offensive tomls

Should hold up well enough to stabilize things until a better path is ready

psionic-k:
I pulled in some drive-by fixes that were strictly necessary to build a slug with musl and `static-web-server`

overlay/mkcrate.nix
```diff
@@ -167,14 +167,20 @@ let
         echo source = \"registry+''${registry}\" >> Cargo.lock
       fi
       mv Cargo.toml Cargo.original.toml
+      # Remarshal version cannot handle escaped double quotes in table key paths
+      while grep '\[[^"]*"[^\\"]*\\"[^\\"]*\\"[^"]*[^]]*\]' Cargo.original.toml; do
+        sed -i -r 's/\[([^"]*)"([^\\"]*)\\"([^\\"]*)\\"([^"]*)"([^]]*)\]/[\1"\2'"'"'\3'"'"'\4"\5]/g' Cargo.original.toml
+      done;
```
edude03, line 173: The PR looks great overall, but it bothers me that I can't look at this bit of code and immediately reason about what it does 🤔 I feel like I'm being nit-picky though and not sure what we could do about it anyway. Any thoughts on possible ways to make it more legible?

psionic-k: Guess I'll draft a comment and explain the hunk at the same time

```shell
# Remarshal was choking on table names of the form:
# [key."cfg(foo = \"a\", bar = \"b\"))".path]
# The regex to find or deconstruct these strings must find, in order, these components:
# open bracket, open quote, open escaped quote, and their closing pairs.
# Because each quoted path can contain multiple quote escape pairs, a loop is employed
# to match the first quote escape, which the sed will replace with a single quote equivalent
# until all escaped quote pairs are replaced.  The grep regex is identical to the sed regex
# but does not destructure the match into groups for restructuring in the replacement.
```

It probably can't be more legible.  This is a problem of writing a regex that has to match a lot of special regex characters ('[' and '\' and ']'.  The only nice thing I could do was consume each inverse up to the match so there's no back reference etc.  Up to writing a parser, I think this about what we can expect.

PR #174, opened 2021-01-20
Support branches for git dependencies

steveeJ:


src/template.rs
```diff
@@ -134,13 +145,21 @@ fn to_source(pkg: &ResolvedPackage<'_>, cwd: &Path) -> Result<Source> {
                 .ok_or(anyhow!("checksum is required for crates.io package {}", id))?,
         }
     } else if id.source_id().is_git() {
+        let branch = if let Some(cargo::core::source::GitReference::Branch(branch)) =
```
edude03, line 148: Minor improvement idea,

why not add `source::GitReference` to `use cargo::core::{...}`

then change this to GitReference::Branch?

src/template.rs
```diff
@@ -90,10 +90,21 @@ pub struct Crate {
 
 #[derive(Debug, Serialize)]
 pub enum Source {
-    CratesIo { sha256: String },
```
psionic-k, line 93: Prefer not to bleed whitespace changes in. Probably disable your auto format. I'll add a rustfmt configuration if we don't have one

:shipit:  Let's merge this, with or without whitespace. Ping me for re-approval if necessary

psionic-k (approved):

src/template.rs
steveeJ: @psionic-k :ping_pong:

psionic-k (approved):
:1st_place_medal:

PR #176, opened 2021-02-06
Flakes support

### LINKED ISSUE #163: Flake support
- [ ] Build as a flake output #161 
- [ ] Include flake in project
- [ ] Equip CI with nix version with flake support to maintain flakes

onsails:
resolves #163


psionic-k:
@onsails got it.  Will review ASAP

psionic-k:
```
From github.com:cargo2nix/cargo2nix
 * [new ref]         refs/pull/176/head -> origin/pull/176
Checking out branch
fatal: reference is not a tree: 29956a0cb02dbecd9f95be0c05fd67032b552828

exit status 128

CircleCI received exit code 128
```
Will check up on this tomorrow to be sure the feedback from CI is coherent with the statuses on this page

examples-flakes/1-hello-world/flake.nix
```diff
@@ -0,0 +1,44 @@
+{
+  inputs = {
+    nixpkgs-mozilla = { url = github:mozilla/nixpkgs-mozilla; flake = false; };
+    cargo2nix.url = github:onsails/cargo2nix/flake;
```
onsails, line 4: note it's my fork here for now to let CI pass

README.md
```diff
@@ -48,15 +47,34 @@ down to the following steps:
...
 ### Declarative debug & development shell
 
 You can load a `nix-shell` for any crate derivation in the dependency tree. The
 advantage of this shell is that in this environment users can develop their
 crates and be sure that their crates builds in the same way that `cargo2nix`
 overlay will build them.
 
-To do this, first run `nix-shell -A 'rustPkgs.<registry>.<crate>."x.y.z"'
-default.nix`.  For instance, the following command being invoked in this
+To do this, first run `nix-shell -A 'rustPkgs.<registry>.<crate>."x.y.z"' default.nix`. For instance, the following command being invoked in this
```
psionic-k, line 77: Is there a `nix dev` equivalent to also mention?

onsails: not that I know of

examples-flakes/1-hello-world/flake.nix
psionic-k: Regarding the fork reference for CI, can we not use the local source as a flake input? I don't want to have to bump flake inputs after merges. Currently, the overlay version in use to test build is from the repo ref being tested, and referencing it via external URL seems to break that

psionic-k:
Considering switching to github actions if the status updates from circle don't get more consistent

I like that alternative overlay.  Had some issues that need cleaning up with the moz overlay as it was

Still considering how best to reconcile flakes with the plan to embed default pins into the Cargo.nix.  Templates might help project setup, but the disparity between flakes & non-flakes can't be hidden it seems

examples-flakes/1-hello-world/flake.nix
onsails: unfortunately, flakes don't support relative input paths [yet](https://github.com/NixOS/nix/issues/3978)

jakeisnt:
Don't have a ton of Nix experience, but I'm using this branch [here](https://github.com/jakeisnt/rustlings-nixl) and it's working well! Thanks for this!

examples-flakes/1-hello-world/flake.nix
onsails: latest nix supports relative paths.
I've force-pushed changes without reference to the fork.

README.md
```diff
@@ -56,10 +56,12 @@ down to the following steps:
    Cargo workspace.
 2. Create a `flake.nix` file with nixpkgs, [cargo2nix] and [nixpkgs-mozilla] inputs
```
psionic-k, line 57: #178 switches to Oxalica's overlay by default.   Want me to merge the changes into this branch?

I appreciate the flake example.

I can fix up the CI branch pointer and merge this if you're ready to have it all in master.

onsails: Sure! I also successfully use this branch in my projects. Let's dot it.

onsails: Sure! But I think that there should be an option to use nixpkgs-mozilla, because in some cases relying on third-party sha256 is not an option. Given that, I thing having `alternative-overlay` example using nixpkgs-mozilla should be enough.

psionic-k: Considering what would count as "some cases."  What can the moz overlay do that the Oxalica one can't, because the github actions are likely scheduled right after rustup updates.  Can moz provide a toolchain that rustup cannot?

onsails: The point of moz for such cases is that you don't rely on third-party provided sha256. AFAIK, with oxalica's you can't force your own sha256 check. However, this check can be implemented in the Oxalica's overlay.

psionic-k: I'll put the argument back in.  We can't pass it into the Oxalica overlay though.  Moz will be available but unmaintained.

In reality it's all trust-on-first-use and riding on top of Rust's content verification in the end.  Oxalica overlay might be better off in nix community repos.  Obviously anyone can pin and anyone can verify the sha256 manually and independently.

edude03:
So, I'd like to keep this moving forward, as flakes are getting more and more used by the nix community (I'm using them as well). 

@onsails, I don't disagree we should keep the ability to use the mozilla nixpkgs overlay, but as we've decided that we want to use the Oxalica, would you be willing to update your PR to use it? Another good reason to do so is Oxalica's overlay already supports flakes out of the box, which is in line with what you're trying to do in this PR I think :) 

I think in a follow up PR we can address the breakage of the mozilla overlay, and as @psionic-k, it'll be considered unsupported (though PRs are very welcome if you're willing/able to add the ability to switch overlays and a circle test for it)

psionic-k:
I've started work on merging the network.  Currently I'm updating to 0.57 to get 2021 edition support.

One thing I want to figure out before merging to master is how to do the `nix-shell` workflow shown here:

https://github.com/cargo2nix/cargo2nix#declarative-debug--development-shell

After updating to flakes, I don't know how to enter the same environment used in that dependency debugging workflow. 

I get a hangup with 0.57 in the nix build where it starts complaining about use of experimental features.  Regular build with rust was fine.  I'm guessing the replacement toml used to build cargo as a nix dependency causes the failure.

psionic-k:
I think I grabbed everything that was in this PR.  I did not attempt to do anything with sha256 or nixpkgs-mozilla.  It looks like there's enough support the oxalica overlay for nightly (although there is another overlay I spotted yesterday).

It might make sense to allow overriding the entire `rustChannel` in overlay/mkcrate/simplified.nix.  The way we're expecting a string and evaluating the rust toolchain eliminates too many options for how to provide rust.

I think this topic needs revisiting in an issue

PR #161, opened 2020-12-16
rustCrateSha256 for compatibility with flakes

onsails:
When a package is built by flakes which require reproducible inputs including downloads via `fetchUrl` cargo2nix fails to build mozilla rust-overlay.
This PR introduces an option rustCrateSha256 which is being passed to rust-overlay to fix the problem mentioned above.

PR #178, opened 2021-03-23
Switch rust overlay

psionic-k:
https://github.com/oxalica/rust-overlay  Said to be a better Rust overlay.  Does seem more focused on a flake-centric world, and flake adoption is being developed in #176  and elsewhere

Tossing this into CI to see what happens


psionic-k:
@onsails  you may find this PR interesting.   Let's reflect on #161  @trha @ebkalderon

> shanesveller — 01/22/2021
> @psionik  I made an extremely casual adaption of your upstream docs to package rust-analyzer in a flake with tunable flake inputs and a fixed version of rust toolchain: https://github.com/shanesveller/flake-rust-analyzer/blob/d8d2b2e8c6137f328edd9021bb0d4011d139762e/flake.nix#L35-L53. while it's perhaps verbose I actually appreciated each opportunity to have full control over the inputs and would hope to see those continue to work (with any reasonable adjustments needed) after any efforts to streamline.

And let's think about how to handle #176 

The big question I have is what behavior do we want Rust projects to be able to use when they switch to flakes?  Building cargo2nix via flakes will just make a few toolchain installations more convenient.  The downstream projects should have a lot more consideration.  The question is what are the downstream projects' likely concerns?

In the discussion to streamline Cargo.nix to include some default pins through some mechanism, it should be expected that these pins are listed as a flake input and therefore can be overridden via flakes mechanism by a project consuming the cargo2nix overlay.  A more monolithic Cargo.nix then would also have to provide some flake information or API hooks

overlay/make-package-set/simplified.nix
```diff
@@ -6,7 +6,6 @@
 }:
 args@{
   rustChannel,
-  rustChannelSha256 ? null,
```
ebkalderon, line 9: It seems that this argument is rendered unnecessary with @oxalica's Rust overlay, as mentioned in commit https://github.com/cargo2nix/cargo2nix/commit/5cd8d79d407894e46735d81ab84896d2e3f5ee5c. If this is the case, why don't we remove all traces of it from the project? If we're already making breaking changes to the public API by switching overlays, forcing users to upgrade and/or regenerate Nix expressions, why keep a legacy attribute around? Is it required anywhere else?

ebkalderon (changes requested):
I really like this PR so far, @psionic-k! :heart: I've left some suggestions and feedback to address some low-hanging fruit. With that said, I think we need to continue to ponder over the questions you've outlined above regarding flake support and streamlining the API ergonomics. I agree that there are some unanswered questions that need to be addressed.

overlay/make-package-set/simplified.nix
psionic-k: I went ahead and removed all instances in the same amended commit

No remaining or foreseen use

ebkalderon (approved):
Looks better to me! Thanks for addressing the changes, @psionic-k. :rocket:

PR #180, opened 2021-03-30
Fix workspace crate paths

### LINKED ISSUE #169: Local patterns defaults frequently omits workspace crates
Setting local patterns really shouldn't be necessary for crates, which are mentioned as workspace members.  When creating `cargo2nix` any files that the `Cargo.nix` will refer to should just automatically be included.

 `   localPatterns = [ ''^(src|tests|hello-deps|assets|templates)(/.*)?'' ''[^/]*\.(rs|toml)$'' ];`

What we really want more likely is simple `.gitignore` behavior and at a minimum omitting target directories where active `cargo` local development workflows will bloat / break caching of the source directory.

Also `localPatterns` doesn't currently work for git resources, which aren't in the store and can't be filtered the same way.  We need a better approach all around.

Something that might work is going along with the `flakes` style assumption that if the repo is a git repo, we fetch the local path as a git resource, thereby honoring `.gitignore` but if the path i a true non-git local path, filtering `target`

A blacklist would also likely be more effective than the current white list.

psionic-k:
Fixes #169

`localPatterns` is used whenever the source is local (not for `fetchGit` etc).  When concatenating crate paths in generated `Cargo.nix`, the `fetchCrateLocal` call filtered each workspace crate to nothing, causing nix to error when it couldn't "unpack the archive" and requiring working around by setting `localPatterns` to include each workspace crate.

The application point for `localPatterns`
https://github.com/cargo2nix/cargo2nix/blob/master/overlay/make-package-set/full.nix#L53-L55

Tested by converting a `fetchGit` to relative path for Rust-Analyzer in 4th example.  Reproduced and fixed issue.  Cargo2nix and other examples don't do this path concatenation and were unaffected.  Git sources are unaffected.

### DIFF
--- a/examples/4-independent-packaging/Cargo.nix
+++ b/examples/4-independent-packaging/Cargo.nix
@@ -212,7 +212,7 @@ in
     name = "base_db";
     version = "0.0.0";
     registry = "unknown";
-    src = fetchCrateLocal workspaceSrc + "/crates/base_db";
+    src = fetchCrateLocal (workspaceSrc + "/crates/base_db");
     dependencies = {
       cfg = rustPackages."unknown".cfg."0.0.0" { inherit profileName; };
       profile = rustPackages."unknown".profile."0.0.0" { inherit profileName; };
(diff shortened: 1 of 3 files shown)


trha (approved):
Nice catch!

PR #185, opened 2021-04-12
Fix missing benchmark dependencies

ollie-etl:
Benchmarks were missing `dev-dependencies`


ollie-etl:
@ebkalderon @edude03 Would it be possible to review: this PR is trivial, but is fixing an existing bug.

edude03 (approved):

PR #189, opened 2021-04-25
Add cross compilation tests

edude03:
It seems like some combo of nixpkgs x cargo2nix x rust doesn't work with musl64 anymore, so I'm going to add some tests to automate tracking down the issue


ebkalderon:
I just found this on Reddit: [PSA - libz-sys on musl no longer links statically by default (unless cross-compiling)](https://www.reddit.com/r/rust/comments/n43pcb/psa_libzsys_on_musl_no_longer_links_statically_by/) Seems there have been quite a few breaking changes to the `musl` compile target in recent versions of Rust.

psionic-k:
Hope the musl situation is improved.  I will explore this while adding a thorough cross compiling example to show how to generate the package set over the supported matrix

PR #195, opened 2021-08-05
upgrade cargo

edude03:
Upgraded cargo because I needed a newer version to support the `resolves = 2` feature.


ebkalderon (approved):
Seems like this should work to unblock users! :+1: Have we verified that the caching behavior of `cargo2nix` behaves correctly under the new resolver, given the assumptions built into the CLI tool and overlay that assume feature resolver 1.0? ([see The Cargo Book](https://doc.rust-lang.org/cargo/reference/resolver.html#feature-resolver-version-2)).

edude03:
I honestly haven’t other than using this PR to build Apollo Rover which uses resolver 2. Since we don’t have great testing infra I think we’ll have to rely on the community to report any bugs they find with 2

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

Issue #184, opened 2021-04-09
Non repeatable build / broken dependency propogation


ollie-etl:
I'm not sure exactly whats up here, but cargo2nix appears broken.

I've been building in CI, and running to successful completion. Getting hold of master and issuing nix-build pulls everything from the binary cache, which no build, as expected.

However:
```bash
echo " " >> ./crates/xrfdc/Cargo.toml
cargo clean
cargo generate-lockfile
cargo2nix -f
nix-build .
```
results in the error below. Cargo is presumably unhappy with the metadata within `encoding_rs`.

I haven't bottomed out the cause yet (otherwise there would be an accompanying PR).

The fact that this has built successfully indicates there is an ordering type bug at play here? Or that cargo is emitting metadeta which renders the build invalid

```json
   Compiling async-graphql v2.8.2 (/build/async-graphql-2.8.2)
{
    "reason": "compiler-message",
    "package_id": "xrfdc 0.1.0 (path+file:///build/source)",
    "target": {
        "kind": [
            "lib"
        ],
        "crate_types": [
            "lib"
        ],
        "name": "xrfdc",
        "src_path": "/build/source/src/lib.rs",
        "edition": "2018",
        "doc": true,
        "doctest": true,
        "test": true
    },
    "message": {
        "rendered": "error[E0460]: found possibly newer version of crate `encoding_rs` which `async_graphql` depends on
 --> src/schema.rs:2:5
  |
2 | use async_graphql::*;
  |     ^^^^^^^^^^^^^
  |
  = note: perhaps that crate needs to be recompiled?
  = note: the following crate versions were found:
          crate `encoding_rs`: /nix/store/md0bfvsf39ch8vki443wsib7sc44vcjd-crate-encoding_rs-0.8.28/lib/libencoding_rs.rlib
          crate `async_graphql`: /nix/store/akgk0qdl4fzn32mhpg54zwjgxh7nzc9a-crate-async-graphql-2.8.1/lib/libasync_graphql.rlib

",
        "children": [
            {
                "children": [],
                "code": null,
                "level": "note",
                "message": "perhaps that crate needs to be recompiled?",
                "rendered": null,
                "spans": []
            },
            {
                "children": [],
                "code": null,
                "level": "note",
                "message": "the following crate versions were found:
crate `encoding_rs`: /nix/store/md0bfvsf39ch8vki443wsib7sc44vcjd-crate-encoding_rs-0.8.28/lib/libencoding_rs.rlib
[...]

PR #197, opened 2021-08-12
Configure cargo

ollie-etl:


ollie-etl:
Sorry, wrong base repo. Its all a bit high pressure currently due to #184

Issue #129, opened 2020-07-14
Build dependencies are built with hostPlatformCpu and hostPlatformFeatures


philipstears:
We had an interesting issue related to my PR https://github.com/tenx-tech/cargo2nix/pull/119.

Everything built fine on our own build servers, and we got a nice CPU optimized asset out of the remote end. But then a customer tried to build it on their own machine and the build failed with SIGILL.

It transpires, that we've got the following dependency chain: `our code -> clap -> indexmap`, and it is the compilation of `indexmap` that was failing.

After some investigation, it turns out that `indexmap` uses a crate called `autocfg` as part of its build script:

https://github.com/bluss/indexmap/blob/master/build.rs#L4

So although the build script gets built with a different set of flags to crates, `autocfg` is just a regular crate and gets built with the same `hostPlatformCpu` and `hostPlatformFeatures` as any other crate, which then meant that the build script binary couldn't be run on the customer's build server because it didn't have those instruction sets.

I figured that was something worth having a conversation about.

One potential solution I could see is to have crates have both build and host variants in `Cargo.nix` and then having the buildDependencies refer to the build variant instead of the host variant.

I apologize if an issue isn't the right place to discuss this, if there's a better place, I'll happily discuss it there :)

ebkalderon:
Interesting find, @philipstears! I wonder if this was what @trha was referring to in the original PR discussion regarding host vs. build platform CPU/features. 😄 Can you confirm if this is the case, @trha?

We are holding off on releasing 0.9.0 for now since there are a couple open PRs that also need review, so there is some time to merge in further changes and fixes to the `target-cpu` integration.

pcpthm:
I was cross-compiling to a different system (x86_64 -> aarch64) and I also had a similar issue of build-dependency not compiling. I have solved my issue with this code:

```nix
let mk = pkgs:
pkgs.rustBuilder.makePackageSet {
    inherit packageFun;
    inherit (pkgs.buildPackages) cargo rustc;
    buildRustPackages = mk pkgs.pkgsBuildBuild;
}; in mk pkgs
```

[...]

PR #199, opened 2021-09-06
Allow direct configuration of target

sphw:
This PR allows users to directly configure what target to build a Rust crate for. In the current version of `cargo2nix` you can cross-compile crates by using a Nix crossSystem, but that only lets you specify target types that are supported by Nix. At first this might sound like a contrived limitation, but there is a real use case for this, SGX builds. The Fortanix EDP uses a tier 3 target called `x86_64-fortanix-unknown-sgx`. Since there isn't a corresponding Nix system, you can't cross compile to the Fortanix ABI.

I solve this problem by allow configuration of build target in `mkCrate` and `makePackageSet`. This allows the user to pass in whatever target they please, with a fallback to the standard `realHostTriple` method. 

I am new to the `cargo2nix` project, so apologizes if I have made these changes in a clumsy manor, or not followed PR etiquette strictly.

### DIFF
--- a/overlay/make-package-set/full.nix
+++ b/overlay/make-package-set/full.nix
@@ -11,6 +11,7 @@
   packageFun,
   workspaceSrc ? null,
   rustChannel,
+  target,
   buildRustPackages ? null,
   localPatterns ? [ ''^(src|tests)(/.*)?'' ''[^/]*\.(rs|toml)$'' ],
   packageOverrides ? rustBuilder.overrides.all,
@@ -45,7 +46,7 @@ lib.fix' (self:
     mkRustCrate' = lib.makeOverridable (callPackage mkRustCrate { inherit rustLib; });
     combinedOverride = builtins.foldl' rustLib.combineOverrides rustLib.nullOverride packageOverrides;
     packageFunWith = { mkRustCrate, buildRustPackages }: lib.fix (rustPackages: packageFun {
-      inherit rustPackages buildRustPackages lib workspaceSrc;
+      inherit rustPackages buildRustPackages lib workspaceSrc target;
       inherit (stdenv) hostPlatform;
       mkRustCrate = rustLib.runOverride combinedOverride mkRustCrate;
       rustLib = rustLib // {
--- a/Cargo.nix
+++ b/Cargo.nix
@@ -11,6 +11,7 @@ args@{
   hostPlatform,
   hostPlatformCpu ? null,
   hostPlatformFeatures ? [],
+  target,
   mkRustCrate,
   rustLib,
   lib,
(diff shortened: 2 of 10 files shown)


overlay/make-package-set/full.nix
```diff
@@ -11,6 +11,7 @@
   packageFun,
   workspaceSrc ? null,
   rustChannel,
+  target,
```
edude03, line 14: @trha or @ebkalderon would be better to review this, but the only thing I notice is that you're not setting a default? 

Seems like this should be `target ? (rustBuilder.rustLib.realHostTriple stdenv.targetPlatform)` or something like that, so that the original behaviour is maintained.

edude03 (changes requested):

overlay/make-package-set/full.nix
psionic-k: The default is handled further upstream in `simplified.nix`

psionic-k (commented):
Will this break existing `Cargo.nix` files?

The `minor` branch is where I'm currently gathering up changes.  The `target` for the rust channel is configured in a really similar way.

I'm likely going to break this into two commits.  That's usually easier to apply when it comes to handling generated changes.

Taking a break. Will review in more detail later.

sphw:
> 
> 
> Will this break existing `Cargo.nix` files?
> 
> The `minor` branch is where I'm currently gathering up changes. The `target` for the rust channel is configured in a really similar way.
> 
> I'm likely going to break this into two commits. That's usually easier to apply when it comes to handling generated changes.
> 
> Taking a break. Will review in more detail later.

This will break existing `Cargo.nix` files. I have made some fixes and changes in our internal branch to this PR that I will push up now. You can see the new commit here: https://github.com/sphw/cargo2nix/commit/8edd83a16b4dc304c23bcee48f77894863c1eab4

psionic-k:
I apologize for the amount of disruption to the master branch.  I'm aiming to get this change in right after releasing 0.9.1.  Recommend creating a new branch and harvesting the non-generated changes and then re-generating.

It will be better to keep the generated files in separate commits in case a re-generation is needed by another change.

psionic-k:
Master is ready to build on top of again.  Sorry for the chaos of all the force pushes.  I will definitely bring this PR in.  I want to address #129 also.

psionic-k:
Changes have been eaten in #214  and tested.   Remembered to add you to the commit author.  Thanks!

I made some modification that enabled old Cargo.nix to be used (if no explicit target is passed) and probably two minor changes to the offsets to enable the `build.rs` to be configured for the correct platform

In the future, not that stdenv.buildPlatform is for build and stdenv.hostPlatform is for the "target".  stdenv.targetPlatform is usually an implementation detail of gcc's that might need to cross compile to make a compiler that runs on a different host and puts out code for a third target

PR #201, opened 2021-10-06
Translating target_arch to nix cpu is required

superboum:
Currently, there is a bug with cargo2nix in the way it interprets `target_arch` conditionals.

`target_arch` values are (partially) defined [in the doc](https://doc.rust-lang.org/reference/conditional-compilation.html#target_arch) as the following set:  x86 x86_64 mips powerpc powerpc64 arm aarch64.

This set does not overlap with how the CPUs are defined by nix in at least two points:
  - rust `x86` refers to `i686` in nix
  - rust `arm` both refers to v6 (identified as `armv6l`) and v7 (identified as `armv7l`) in nix.

This patch adds a function to map `target_arch` to the correct `cpu`.

This patch has been tried with:
  - `ring` - otherwise it fails with `use of undeclared crate or module spin` when compiled with `armv7l-unknown-linux-gnueabihf`
  - `sha2` - otherwise it fails with `use of undeclared crate or module cpuid_bool` when compiled with `i686-unknown-linux-musl`

### DIFF
--- a/src/platform.rs
+++ b/src/platform.rs
@@ -11,6 +11,18 @@ pub fn to_expr(p: &Platform, platform_var: &str) -> BoolExpr {
     }
 }
 
+fn platform_single(platform: &str, nix_cpu: &str) -> BoolExpr {
+    BoolExpr::Single(format!("{}.parsed.cpu.name == {:?}", platform, nix_cpu))
+
+}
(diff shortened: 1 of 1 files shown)


psionic-k (approved):
Sounds good.  Applying.  Thanks!

PR #205, opened 2021-11-10
Minor Version Release

### LINKED ISSUE #158: Put intermediate files into separate dependency
This is the first issue focused on long-term improvements to architecture.

The various files that we write into the outputs in order to support later linking have negative effects on the generation of profiles and also on closure sizes as nix assumes that all these files that we need at link time etc are parts of output closures.

The `bin` target separation in #153  is basically a hack to address the symptom of this issue.

These files could be unique for some builds while having no effect at all on the actual outputs of dependencies, possibly resulting in duplication in the store?

Another issue we could be facing is taking on too much responsibility from cargo that may have been baked into earlier releases when other solutions in cargo meta data etc were not available.  Libraries such as cargo-lock and guppy have come about and may solve some of the issues in a more lightweight manner

psionic-k:
~Just smoke testing the Rust Analyzer flake on CI~

Changlog draft: 
- updated Rust edition support to 2021 & cargo to 0.57
- nix flake
- switched CI to actions
- fixed hostPlatformFeatures
- support for profile-opts and codegen-opts (this likely needs a second pass for architecture.   Note as experimental)
- mac support fixed
- target_arch translation fix

[insert changes since 0.9.0]

This PR will fix #158 well enough to close it.  Turned out issue was some impurity / non-determinism that has gone away

Remaining items to put this into master:

- [x] convert examples to flakes
- [x] fix / verify osx
- [x] remove circle CI and update actions to build all attributes we want to test
- [ ] version bump

### DIFF
--- a/examples/1-hello-world/flake.nix
+++ b/examples/1-hello-world/flake.nix
@@ -1,8 +1,8 @@
 {
   inputs = {
-    cargo2nix.url = "path:../../";
+    # cargo2nix.url = "path:../../";
     # Use the github URL for real packages
-    # cargo2nix.url = "github:cargo2nix/cargo2nix/master";
+    cargo2nix.url = "github:cargo2nix/cargo2nix/master";
     flake-utils.url = "github:numtide/flake-utils";
     rust-overlay.url = "github:oxalica/rust-overlay";
     rust-overlay.inputs.nixpkgs.follows = "nixpkgs";
--- a/examples/2-bigger-project/flake.nix
+++ b/examples/2-bigger-project/flake.nix
@@ -1,8 +1,8 @@
 {
   inputs = {
-    cargo2nix.url = "path:../../";
+    # cargo2nix.url = "path:../../";
     # Use the github URL for real packages
-    # cargo2nix.url = "github:cargo2nix/cargo2nix/master";
+    cargo2nix.url = "github:cargo2nix/cargo2nix/master";
     flake-utils.url = "github:numtide/flake-utils";
     rust-overlay.url = "github:oxalica/rust-overlay";
     rust-overlay.inputs.nixpkgs.follows = "nixpkgs";
--- a/examples/3-static-resources/flake.nix
+++ b/examples/3-static-resources/flake.nix
@@ -1,8 +1,8 @@
 {
   inputs = {
-    cargo2nix.url = "path:../../";
+    # cargo2nix.url = "path:../../";
     # Use the github URL for real packages
-    # cargo2nix.url = "github:cargo2nix/cargo2nix/master";
+    cargo2nix.url = "github:cargo2nix/cargo2nix/master";
     flake-utils.url = "github:numtide/flake-utils";
     rust-overlay.url = "github:oxalica/rust-overlay";
     rust-overlay.inputs.nixpkgs.follows = "nixpkgs";
--- a/examples/4-independent-packaging/flake.nix
+++ b/examples/4-independent-packaging/flake.nix
@@ -1,8 +1,8 @@
 {
   inputs = {
-    cargo2nix.url = "path:../../";
+    # cargo2nix.url = "path:../../";
     # Use the github URL for real packages
-    # cargo2nix.url = "github:cargo2nix/cargo2nix/master";
+    cargo2nix.url = "github:cargo2nix/cargo2nix/master";
     flake-utils.url = "github:numtide/flake-utils";
     rust-overlay.url = "github:oxalica/rust-overlay";
     rust-overlay.inputs.nixpkgs.follows = "nixpkgs";


psionic-k:
@edude03  How did your experiment with Github Actions go?  Any info to pass on before I attempt to do the same switch over?  I have a working actions example from my website source I can try to pull over.  There's a branch in the Github network with some fixes for mac btw.

Preparing to cut a minor release.  There's still a few commits in the Github network I want to pull in.

This version added Edition 2021 support necessary to build Rust Analyzer.  Currently I'm using the build with flakes.

Did manage to find a new issue where somehow both allocators were enabled in the generated Cargo.nix.  It was easy to just remove one of the features by hand fortunately, but I'll have to figure out why Rust Analyzer ends up with both features.  Change was here  https://github.com/cargo2nix/cargo2nix/pull/205/files#diff-af9181ec0d50c28fbf479b93c2e9a07a7725714cfa7c1cc0ca0c6ab708772d83R1803

edude03:
The GitHub action branch I have open works, except the Mac build is failing. Let’s back port the fix then merge it

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

psionic-k:
@edude03 https://github.com/metabreak/cargo2nix/commit/03ace250526fce85d9205d5387aa6a8009b89ae3

Was this the issue you were running into on Mac?   Just got through blowing away my Mac, so not ready to test for a little while.

psionic-k:
https://www.reddit.com/r/NixOS/comments/qqrfq5/from_nixshell_to_nix_develop/hk1xf7w/?context=3

This is the other thing I want to get fixed up.  I don't think we have the option to do `nix develop .#...cargo."0.57.0"` etc like we could pre-flakes.  There are plenty of times I would not have been able to quickly figure out an override without this workflow.

psionic-k:
We have gained a really nice side effect!

https://github.com/cargo2nix/cargo2nix/issues/158

I made the hello-deps example into a flake and then varied the features as I had done before, but the rand crate (which is independent of the feature) generated identical outputs.

```
      "/nix/store/5kc0m5h6rdng9qh2hj4jcjnnxqywijnn-crate-rand-0.8.4.drv": [
        "out"
      ],
```

It is evidently possible to use flakes and obtain more coincidentally identical dependencies, meaning we have useful build caching between distinct derivations.

IMO this was the biggest behavior cargo2nix would have been expected to solve that was not actually solving.

psionic-k:
Aight, how to expose a single dependency crate for `nix develop .#cargo` etc

```nix

      #...
      # in the flake outputs
      in rec {

        packages = {
          inherit rustPkgs;
          package = rustPkgs.workspace.cargo2nix {};

          # exposing one derivation to nix develop
          cargo = rustPkgs."registry+https://github.com/rust-lang/crates.io-index".cargo."0.57.0" {};
        };

        defaultPackage = packages.package;
      }
    );

```

So now I just need to map over these attributes.  Having to call them all feels like something is wrong or I will do it wrong if I do it naively.  What is the function we're returning if it doesn't matter what the arguments are?

This will fix up the dependency debug workflow.  Pretty close to everything needed to ship flakes.

psionic-k:
We're going with this as the recommendation for getting a dependency shell:

```shell
  nix develop --derivation /nix/store/bsl20cjl1k3d4649103hb3lqpn3zppd7-crate-cargo-0.57.0.drv
```

Frankly this is easier and more natural after a dependency fails than trying to figure out the correct attribute paths at  `rustPkgs.<registry>.<name>.<version>`

psionic-k:
@edude03 metabreak@03ace25  including libiconv will fix each darwin dependency.  Since it likely affects all crates built on darwin, I will guess that this was fixed in stdenv by now.  If not, I think we just need to attach it to buildDependencies in mkcrate.nix or possibly by adding something to patches in overrides.nix

Can't recall if actions need to be enabled by blessing the newly introduced yaml somehow :thinking:   Anyway, I was going to try with latest versions before going to deep fixing libiconv everywhere.  Faster to just re-setup my mac.  Go beat me to it!

Remaining items to put this into master:

- [ ] convert examples to flakes
- [ ] fix / verify osx
- [ ] remove circle CI and update actions to build all attributes we want to test
- [ ] version bump

Decided to move the target fixes to next PR

psionic-k:
@edude03 https://github.com/cargo2nix/cargo2nix/pull/205/commits/5b4e87dca80af25558430d483e7a8c59635cd0de

There you go.  Again, not sure if its necessary in latest nixpkgs.  Logic says it should work for CI.

psionic-k:
Our flake seems a bit ghetto as is.  Likely not all commands will work.

Taking some cues off of the [schema](https://nixos.wiki/wiki/Flakes#Output_schema) and some [examples in nixpkgs](https://github.com/NixOS/nixpkgs/blob/342bd7ae59c70366bc36d1bf2a259c864b446421/pkgs/top-level/stage.nix#L157-L165), I think we can make an example using `crossSystem` and providing `pkgsCross` outputs similar to the way that nixpkgs does it.  

[this result](https://github.com/positron-solutions/hello-deps/blob/master/flake.nix) was from me removing flake-utils and using nixpkgs.lib instead.  Was glad to see musl building just fine on Linux.

Also I might be using "follows" incorrectly,..  hello-deps doesn't use flake-utils but I still had tol define it as an input in the flake.

PR #208, opened 2021-11-13
Release 0.10.0

### LINKED ISSUE #158: Put intermediate files into separate dependency

psionic-k:
Hope it looks pretty.  Flakes are really nice.  The new nix commands are a lot of fun.

Issues to open:
- sub-flakes cannot reference relative paths in CI.  Works locally.
- enabled features on Rust Analyzer leads to both allocators being enabled at the same time.  Errata noted.
- cross compiling is good, but we do need an example of mapping over the supported systems & static support like nixpkgs does it

Changlog draft: 
- updated Rust edition support to 2021 & cargo to 0.57
- nix flake
- switched CI to Github actions
- fixed hostPlatformFeatures
- support for profile-opts and codegen-opts (this likely needs a second pass for architecture.   Note as experimental)
- mac support fixed
- target_arch translation fix

[insert changes since 0.9.0]

This PR will fix #158 well enough to close it.  Turned out issue was some impurity / non-determinism that has gone away

Remaining items to put this into master:

- [x] convert examples to flakes
- [x] fix / verify osx
- [x] remove circle CI and update actions to build all attributes we want to test
- [x] version bump
- [x] fix cachix filter path issue

Original PR was here:
https://github.com/cargo2nix/cargo2nix/pull/205


psionic-k:
```
       > /nix/store/dban9r73fqzzd04bz6p7xn8v8p3s657z-rustc/bin/rustc: line 19: /nix/store/mzqqb6lqkpx33wml4bqsf6zsamjifcyy-rust-minimal-1.56.1/bin/rustc: Argument list too long
```

Guessing our epic argument lists grew a bit.  How to up the stack size on Mac?   I'm going to have to get my Mac set back up for development unless one of you beats me to this.  It's possible that the argument list is getting filled up with a lot of duplicates or unnecessary flags.

Probably just merge it.  The most important jobs are already working.  Mac build & Linux build.

psionic-k:
image

You guys, I think I'm narrowing in on a fix lol

psionic-k:
RustAnalyzer on mac now attempts to build (argument list too long is fixed) but then linking fails with a lot of duplicate symbol warnings.  

Won't be debugging this until I update my mac.  Going to take too long to round trip this in CI.  Cargo2nix itself is fine and most simpler crates will build no problem.  I think this branch is ready to use for anyone interested.

PR #214, opened 2021-11-14
Host Platform build.rs for Cross Build

### LINKED ISSUE #36: Improve localPatterns
Suggest splitting this:

```
    localPatterns = [
      ''^(src)(/.*)?''
      ''^(migrations)(/.*)?''
      ''^(proto)(/.*)?''
      ''[^/]*\.(rs|toml)$''
    ];
```

...into two lists of
`includeLocalFiles`
`includeLocalDirs`

### LINKED ISSUE #129: Build dependencies are built with hostPlatformCpu and hostPlatformFeatures

### LINKED ISSUE #202: dangling path
hi!

thx for this project, it has been very helpful for scaffolding my projects 🙏

I noticed what seems to be a dangling path in the overlay here: https://github.com/cargo2nix/cargo2nix/blob/381f9760d471bc07925d13ac924b0a1bb86d78a5/overlay/default.nix

see example to reproduce:

```
vayu :: /tmp/cargo2nix-test » bat flake.nix
───────┬──────────────────────────────────────────────────────────────────────────────────────────────
       │ File: flake.nix
───────┼──────────────────────────────────────────────────────────────────────────────────────────────
   1   │ {
   2   │   description = "cargo2nix-test";
   3   │
   4   │   inputs = {
   5   │     nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
   6   │     flake-utils.url = "github:numtide/flake-utils";
   7   │     cargo2nix.url = "github:cargo2nix/cargo2nix/feature/add-basic-flake-support";
   8   │   };
   9   │
  10   │   outputs = { self, nixpkgs, flake-utils, cargo2nix, ... }:
  11   │     flake-utils.lib.eachDefaultSystem (system:
  12   │       let
  13   │         overlays = [
  14   │           (import "${cargo2nix}/overlay")
  15   │         ];
  16   │         pkgs = import nixpkgs {
  17   │           inherit system overlays;
  18   │         };
  19   │       in
  20   │       {
  21   │         defaultPackage =
  22   │           pkgs.rustBuilder.mkCrate;
  23   │       });
  24   │ }
───────┴──────────────────────────────────────────────────────────────────────────────────────────────
vayu :: /tmp/cargo2nix-test » nix build
error: getting status of '/nix/store/jk4sv25dglwi8j0d7bwxn0f7npshp7jq-source/overlay/crate.nix': No such file or directory
(use '--show-trace' to show detailed location information)
```

I'm using the flakes branch but it appears on `master` too.

### LINKED ISSUE #203: targeting `wasm32-unknown-unknown`
hi!

thx for this project, it has been very helpful for scaffolding my projects :pray:

I am wondering if it's possible to target wasm using this project, perhaps via `makePackageSet*`.

I have a project I currently build with `wasm-pack` / `cargo build --target wasm32-unknown-unknown`, and I would like to be able to `nix build` it reproducibly.

I saw [this issue](https://github.com/cargo2nix/cargo2nix/pull/6) which makes me hopeful this is possible, but I haven't had time to investigate the internals of this project sufficiently to understand what is possible/doable/difficult.

any thoughts or pointers would be much appreciated.

this is of personal & work-related interest to me and I am up for implementing a solution or helping in an effort.

thx :slightly_smiling_face:

psionic-k:
This PR strips out the `localFilters` because:

1. it's useless for independent sources
2. frequently a source of breakage for users
3. flakes already respect .gitignore
4. the more likely use pattern people wanted all along is to filter something out (which can be done during unpack) rather than to select something via regex

I used the static resources example and just built with cargo to get a target dir.  Zero effect on the resutls of nix builds.  `target` did not need to be filtered from sources.  Flakes :partying_face: 

## So Example 3 is Now Cross-Compiling!

The third example's use of a build.rs script is perfect for debugging the host and build platform offsets.  We will fix everything.  :fist_oncoming: 

The cross compiling example with wasi linking can now be compiled on Linux and run with `wasmtime`

Lets fix....

- [x] fix #203
- [x] close #199 
- [x] fix #129
- [x] fix #36 (local sources already dead)
- [x] fix #202 (drive by fixup)
- [x] split test matrix for darwin & linux to get Rust Analyzer back into CI


psionic-k:
This PR should be added to with linear, small commits that can later be cleaned up by squashing in rebase.  Let's all play nice!

psionic-k:
Kind of expecting at least something would break before I even got to building the packages matrix across sytems & crossSystems.

aarch64-crossplatform compile on Linux system was successful:

```
./result-bin/bin/cross-compiling 
bash: ./result-bin/bin/cross-compiling: cannot execute binary file: Exec format error

nix shell nixpkgs#qemu

qemu-aarch64 ./result-bin/bin/cross-compiling 
Hello world

file ./result-bin/bin/cross-compiling 
./result-bin/bin/cross-compiling: ELF 64-bit LSB pie executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /nix/store/31pjpb2jfj5jqb3hwk2s5xmv7zqi1xf2-glibc-aarch64-unknown-linux-gnu-2.32-54/lib/ld-linux-aarch64.so.1, for GNU/Linux 2.6.32, not stripped
```

Kind of disappointed that there's nothing to start fixing for host offsets to get the build.rs working

psionic-k:
Latest changes add explicit target configuration.

I think we need to split up the `target` for selecting a Rust toolchain and the target configuration for the code output, if these are not the same.  Judging by the weird vendor targets people brought up elsewhere, I'm convinced that the Rust toolchain selected by one of its targets doesn't always match what code that toolchain will be used to generate.  We can mostly handle this for Rust crates without much extra work I expect.

For C deps, there may be more work in the form of providing a linker or some lib without bringing in a whole stdenv.  

`buildPhase` overrides for `mkcrate.nix` can probably handle any extra changes necessary for outputing code or linking that go beyond merely passing options to `rustc`

psionic-k:
There's two places we translate platforms and host triples.  `/overlay/lib/real-host-triple.nix` and `/src/platform.rs`   Both need some maintenance many targets were added.

I'm working on web assembly right now.

psionic-k:
https://github.com/oxalica/rust-overlay/blob/f0a37690fae87b08c3e42c14e209d78fa164d820/rust-overlay.nix#L99-L107

This hunk in the Rust Overlay is somehow resulting in an attempt to get a Rust for the host platform.  We only need build platform.  Going to fork and PR master if I can't find a configuration route to get the correct Rust

psionic-k:
Current status is really close.  Building with explicit target for `wasm32-wasi` gets about 95% of the way through the build and then fails because stdenv isn't set up for wasi.  If I set `crossSytstem` for wasi, the oxalica overlay suddenly doesn't find all the components that it wants.

psionic-k:
Evidently there is now a `toRustTarget` function in nixpkgs https://github.com/oxalica/rust-overlay/search?q=toRustTarget

psionic-k:
https://github.com/oxalica/rust-overlay/pull/58    I've updated the oxalica overlay with what I think are the correct offsets.

If you comment out the aarch64 and uncomment the wasi `target` and `crossPlatform` then on Linux you get a wasi build that is extremely close to finishing and just needs to help figuring out how to link.  I think we have to differentiate the `mkcrate` for crossSystem a little more effectively.  Options like `useLLVM` are definitely not what we're used to.

psionic-k:
Because the cross example has been observed to succeed on Darwin for aarch64, this branch is going to start stabilizing.   History will be overwritten with force pushes.  I will cut the rust analyzer example on darwin if I have to.  IIRC it won't be easy to debug until I switch back to a mac for a bit.

:x:  Wasm will not make this PR because it evidently is totally working for linking.

:heavy_check_mark: I will look into .cargo/config.toml again to see if it can be incorporated without any tricky plumbing

psionic-k:
Github rate limiting is only going to get worse.  No fun.

psionic-k:
This is on hold for about a week or more for anyone who wants to jump in

Issue #157, opened 2020-12-09
Embed overlay pin into Cargo.nix


psionic-k:
There's little reason to require the user to supply their own overlay pins while baking the Cargo.nix.  It adds a lot of process overhead to simple packaging and in 99% of cases the right answer is to use the same overlay version as at the time of creating the Cargo.nix

psionic-k:
Just ran into an issue with an odd version of the nixpkgs-mozilla overlay giving me a rustc that had missed libz.so.1 in it's output.  Reinforces the need to pin during Cargo.nix generation and default to the same pins as the version of Cargo2nix.  The error initially looked like I needed a native dep and only after shelling in was it apparent that rustc was broken

Sample error message:
```
[nix-shell:~/debugging/yansi-term-0.1.2]$ /nix/store/qzrglv2ys3k4549ykx3ymyig9rykb241-rust-1.48.0-2020-11-16-7eac88abb/bin/rustc --version
/nix/store/qzrglv2ys3k4549ykx3ymyig9rykb241-rust-1.48.0-2020-11-16-7eac88abb/bin/rustc: error while loading shared libraries: libz.so.1: cannot open shared object file: No such file or directory
```

psionic-k:
We can go several steps farther here in terms of burying the noise for most projects

* Default to placing the Cargo.nix into the /nix directory like niv.  For multiple-workspace shells, it's helpful
* Rename Cargo.nix to match the workspace name, becoming `myPackageWorkspace.nix`
* Embed default pins to avoid `packageFun` to `Cargo.nix` & `nixpkgs-mozilla` version mismatch pitfalls
* Construct the package set with all default args
* Provide the arguments to `mkShell` via function on the workspace so the user doesn't have to construct it manually

The result would be `(callPackage nix/static-web-serverWorkspace.nix { }).static-web-server.bin` to for instance, obtain the static web server program.

If workspaces can expose their shell dependencies in a set that can be merged, then we can build a shell by consuming several workspaces:
```nix
{
  sources ? import ./nix/sources.nix,
  system ? builtins.currentSystem,
  crossSystem ? null, #(import sources.nixpkgs {}).lib.systems.examples.musl64,
}:
let
  # get our pins in scope for callPackage
  inherit (sources) cargo2nix nixpkgs nixpkgs-mozilla;
in let
  staticServerWorkspace = callPackage nix/static-web-serverWorkspace.nix {};
[...]

PR #224, opened 2021-12-05
Be friendly to users

### LINKED ISSUE #156: Honor Cargo.toml "links" Attribute
Probably 95% of crate dependency failures on Linux are due to missing native deps that are trivially identifiable via the Cargo toml `links` field.  Build.rs crates frequently attempt to build a library and an output might be yielded, but it's safe to say that the nixpkgs libs are almost always the better choice for reproducibility if only that.  The link issue on #153 on mac didn't at first appear to be the type of issue that a simple override fixes, but it was, and this is very painful for users, manifesting as tricky failures to link rather than a compiler outright missing native sources.

https://doc.rust-lang.org/cargo/reference/manifest.html#the-links-field

I'm guessing most people don't want Cargo2nix for it's capability to do what cargo does but for it's potential to simplify / normalize the native deps issues that the `links` field was meant to help solve.

We can and should search for a package corresponding to `links` field declarations and just include that native dep automatically.  Overrides in the overlay can then be opt-out

### LINKED ISSUE #166: Use with nixpkgs rustc?
The examples here all use `nixpkgs-mozilla`. I'd prefer to use the rustc provided by Nixpkgs, if possible—can that be done? If not, what needs to happen for it to work?

This would presumably be a requirement for https://github.com/cargo2nix/cargo2nix/issues/42.

### LINKED ISSUE #173: Workspace Path Reference Breakage
https://github.com/getzola/zola/blob/master/components/config/src/highlighting.rs#L9-L17

```rust
lazy_static! {
    pub static ref SYNTAX_SET: SyntaxSet = {
        let ss: SyntaxSet =
            from_binary(include_bytes!("../../../sublime/syntaxes/newlines.packdump"));
        ss
    };
    pub static ref THEME_SET: ThemeSet =
        from_binary(include_bytes!("../../../sublime/themes/all.themedump"));
}
```

This creates a path dependency from the config crate to the workspace, but we cannot accurately express this dependency yet.

One possible solution is to include the entire workspace source and modify the build commands to account for the directory shift.  Cargo must be prevented from attempting to rediscover the workspace in this case, likely adding a `[workspace]` table to the temporary TOML

Other more fragile workarounds are available, such as patching the source paths and including the files explicitly

### LINKED ISSUE #183: Nondeterministic list order in Cargo.nix
A case was discovered where cargo2nix would generate Cargo.nix output with differing list orders.  There's no expected consequence to repeatable evaluation, but it's a reasonable expectation to maintain bijection between the Cargo.lock and the Cargo.nix. 

Reported by @TylerRichie

### LINKED ISSUE #184: Non repeatable build / broken dependency propogation

### LINKED ISSUE #211: Features may not match those enabled from workspace
Rust Analyzer winds up with both mimalloc and jemalloc enabled.  This was patched out by hand.  The greater problem is that features vanishing or appearing when switching from a direct cargo build to a nix build.  Features are very common, so this is very high priority.

### LINKED ISSUE #213: Re-enable Rust Analyzer build
Had a failure on Mac due to some duplicate symbols during linking.  jemalloc related.  My mac is not set up for development right now and would like for someone to jump in and debug the build.  Probably just adjust the nixpkgs rev or make a crate override and done.

If you get it working, please PR and re-enable the CI for Rust Analyzer in .github/workfkows/ci.yml

### LINKED ISSUE #223: Provide proper overlay(s) from flake
It seems as if the `overlay` and `overlays` outputs have an extra layer for the `system`.

As you can see that is not required for ooverlays:

https://nixos.wiki/wiki/Flakes#Output_schema

### LINKED ISSUE #226: rust nightly - error: attribute 'nightly-2019-12-05' missing
to build [c2rust](https://github.com/immunant/c2rust), i must pin the rust version to rust [nightly-2019-12-05](https://github.com/immunant/c2rust/blob/master/rust-toolchain)

naive solution

```nix
# flake.nix
{
          #rustChannel = "1.56.1";
          rustChannel = "nightly-2019-12-05"; # error: attribute 'nightly-2019-12-05' missing
          rustChannel = "nightly"; # error: attribute 'nightly' missing
```

https://github.com/cargo2nix/cargo2nix/blob/ae8a5e699f57b446f9507b97a3c97a4e9fe06aa3/overlay/make-package-set/simplified.nix#L17

<details>
<summary>default.nix with buildRustPackage</summary>

this works by creating one large derivation, but i want many small derivations, hence cargo2nix

```nix
/*
nix-build -E 'with import <nixpkgs> {}; callPackage ./default.nix {}'
*/

/*
usage
  to generate compile_commands.json, use
  nix-shell -p python3Packages.compiledb
  cd your_c_project && ./configure && compiledb -n make
*/

{ lib, pkgs, fetchFromGitHub, callPackage, makeRustPlatform }:

let

  # use custom rust version
  # fix: error[E0462]: found staticlib `rustc` instead of rlib or dylib
  # https://github.com/immunant/c2rust/issues/254#issuecomment-624631037
  # https://github.com/immunant/c2rust/blob/master/rust-toolchain

  rustPlatform = mkRustPlatform {
    date = "2019-12-05";
    channel = "nightly";
  };

  mkRustPlatform = { date, channel }:
    let
      mozillaOverlay = fetchFromGitHub {
        owner = "mozilla";
        repo = "nixpkgs-mozilla";
        rev = "cbc7435f5b0b3d17b16fb1d20cf7b616eec5e093";
        sha256 = "LKqAcdL+woWeYajs02bDQ7q8rsqgXuzhC354NoRaV80=";
      };
      mozilla = callPackage "${mozillaOverlay.out}/package-set.nix" {};
      rustSpecific = (mozilla.rustChannelOf { inherit date channel; }).rust;
    in
    makeRustPlatform {
      cargo = rustSpecific;
      rustc = rustSpecific;
    };

in

rustPlatform.buildRustPackage rec {
  pname = "c2rust";
  version = "2021-08-29";

  src = ./git/c2rust;

  buildType = "debug"; # build faster

  /*
  src = fetchFromGitHub {
    owner = "immunant";
    repo = pname;
    rev = "9d6f830ee7672d31e816df1158b4ba4aad81d8c0";
    sha256 = "Wkf1mM6+eUQBf4Zg3fFXGK14jTvCoD3Nk96MVht8pSs=";
  };
  */

  nativeBuildInputs = [
    pkgs.python3 # c2rust-refactor -> python3 process_ast.py
    pkgs.pkgconfig # c2rust-bitfields-derive -> pkg-config --libs --cflags openssl
    pkgs.llvm # c2rust-ast-exporter -> llvm-config
    pkgs.llvm.dev # c2rust-ast-exporter -> #include "llvm/Support/CommandLine.h"
    pkgs.cmake # c2rust-ast-exporter

    pkgs.tinycbor # c2rust-ast-exporter
  ];

  /*
    FIXME c2rust-ast-exporter
    -- Could NOT find Clang (missing: Clang_DIR)
  */

  # copy paste from c2rust/shell.nix
  buildInputs = [
    pkgs.openssl
    pkgs.zlib
    pkgs.clang

    #pkgs.tinycbor # c2rust-ast-exporter
  ];

  cargoSha256 = "FsPHWZisxGPlDIfx3SJrJpC2meNWL3ReJQx6Ha/z1Bc=";
  #cargoSha256 = "0000000000000000000000000000000000000000000000000000";

  cargoBuildFlags = [ "--locked" ];

  meta = with lib; {
    description = "Migrate C code to Rust";
    homepage = "https://github.com/immunant/c2rust";
    license = licenses.bsd3;
    #maintainers = [ maintainers.todo ];
  };
}
```

</details>

### LINKED ISSUE #227: Update to Rust 1.57
:pancakes:

### LINKED ISSUE #231: How to add rustfmt and clippy to shell
I couldn't find any documentation on how to add `clippy` and `rustfmt` to the `workspaceShell`.

### LINKED ISSUE #233: a string that refers to a store path cannot be appended to a path
```
error: a string that refers to a store path cannot be appended to a path

       at /nix/store/qsbymj5zlc7z5fnb9sg0b3kva803mjrl-source/overlay/lib/fetch.nix:16:7:

           15|     in
           16|       /. + builtins.readFile (runCommand "find-crate-${name}-${version}"
             |       ^
           17|         { nativeBuildInputs = [ jq remarshal ]; }
(use '--show-trace' to show detailed location information)
```
The exact copy of code works on my local machine with `nix-2.6.0pre20211217_6e6e998`
The above error is origninated from `nix 2.6.0`. I think we would be better off to use `builtins.toPath` instead of `/.`

### LINKED ISSUE #239: hard to read JSON output
Failing builds are shown in a JSON output which makes visually parsing the error message difficult. Is there an other option or do you pipe the output to something specific ?

psionic-k:
Working on some small upgrades to usability, fixing some issues of presentation

- [x] readme is a mess.  Word salad -> solid points in the order that a curious user wants the information
- [x] expose attributes for import
- [ ] ~#156~  (pushed to next release)
- [x] fix #211 
- [x] provide combined overlay (related to #157)
- [x] fix #166 and fix #226 
- [x] fix #223 
- [x] fix #173 
- [x] fix #213 
- [x] fix #184  (Adding a note.  This issue is Errata on Rust 1.56, not a bug) 
- [x] fix #183 
- [x] fix #227
- [x] fix #231 
- [x] fix #233 
- [x] fix #239  
- [x] removed the requirement for `-f` switch   
- [ ] ~add container build example for something OCI compliant.  Might as well build a Rust k8's operator example~ (next) #230 
- [x] highlight the dev shell!  If teams can't get up to speed locally, they don't care about nix use in CI & CD and it tends to die


psionic-k:
Just pushed a spike for an auto-links behavior.

After getting it into the Cargo.nix, I started to consider how do we want to:

-  provide users with a sensible override path that is at least as easy as the current manual package overrides
-  support different platforms for overrides in a shorter declarative structure
-  interoperate between auto-links and the heavier `makeOverride`, which can do a lot more

### Correct links override structure

The only real question on declarative data structure is if we should map by links and then hostPlatforms or by hostPlatforms and then links.  I think we should do links and then hostPlatforms.  The "default" key should be Linux because it's going to just work more frequently.  This would be followed by `isDarwin` and `isAndroid` etc.  I'll figure something out.

### MakeOverride Interoperation

To interoperate with `makeOverride`, I think I should use the map to generate a `makeOverride` and then let this be applied in the standard fixed point (somewhere deep in [/overlay/make-package-set/full.nix]).  

### Override Order...

The automatically generated overrides will be wrong occasionally.  The user will want their fix override to clobber the generated one or prevent it from being applied at all.  This might be handled easiest by making `overrides.all` have a better interface.  I'll think about it.

### More Convenience for Overrides

I'm then going to provide a convenience function so that the user can _quickly_ and _easily_ augment the overrides in `makePackageSet`.  If you just want to add a links rename override, it should be something like `rustLib.overrideLinks ( linkname [ pkgs.dep1 pkgs.dep2 ] )`  Our existing workflows are unnecessarily verbose and people hate them.

:x:  the examples Cargo.nix files are going to be totally broken for the moment.  Don't look too hard.  I'll test RustAnalyzer tomorrow.

Still checking into if llinks can have multiple keys or be a comma separated list.  Cargo apI makes me think it might be comma separated list, but I've never seen one yet.

Sure I'll figure out some decent answers tomorrow but please opine.

psionic-k:
What links added to the Cargo.nix looks like for cargo2nix

https://github.com/cargo2nix/cargo2nix/pull/224/files#diff-8a24bca64c849bfd62296fbd5ee747ea42da29cbb0983c18e541eb050436b49bR907

psionic-k:
Our existing user-facing API's :rofl:  

```nix
    # ... required arguments not shown
  
    # Use the existing all list of overrides and append your override
    packageOverrides = pkgs: pkgs.rustBuilder.overrides.all ++ [
    
      # parentheses disambiguate each makeOverride call as a single list element
      (pkgs.rustBuilder.rustLib.makeOverride {
          name = "fantasy-zlib-sys";
          overrideAttrs = drv: {
            propagatedNativeBuildInputs = drv.propagatedNativeBuildInputs or [ ] ++ [
              pkgs.zlib.dev
            ];
          };
      })
      
    ];
  };
```

 That is a lot of annoyance just to tack on a single dep.  The only information really needed is a list of `[[linkname platform [ pkgs.foo, pkgs.bar ]]` 

This is kind of what I'm thinking of...

```nix
    #...usual stuff
    linkOverrides = [{linkname x hostPlatform = [ dep1 dep2 ];}];
    extraOverrides = { packageName x hostPlatform = { combiningOverrideKey = [ pkgs.foo ];};};
  };
```

Not sure how to write `packageName x hostPlatform` composite keys yet in nix, but it would make all overrides have a consistent form independent of if Darwin and Linux etc have different overrides or not.

psionic-k:
Got rid of the `rust-overlay` management from the user perspective.  An ambitious user can still provide the overlays in order manually if absolutely necessary.

This brings me back to the conversation about wanting to embed the overlay information directly into the Cargo.nix.

Cargo.nix could itself become a flake, which would be nice for overriding pins, but then we would be limited to exporting the flake schema attributes to play nice.  https://nixos.wiki/wiki/Flakes#Output_schema   It raises the question of where the convenience functions would go.

Perhaps it is inevitable that to get some supplemental `rustLib` functions.  There's not much size advantage to the flake.nix by not setting up nixpkgs.

The great evil being chased is when the Cargo.nix and overlay version are mismatched.  Should we include an overlay flake separately? In a sub-directory?  #157   Cargo.nix itself is just a function, and this doesn't really have a proper place in a flake output unless it's as an overlay.  It would be obtuse to make Cargo.nix into an overlay that provides only the packageFun.

edude03:
Just fyi, I've looked this over, but would be nice to get @ebkalderon and @trha in on API changes

psionic-k:
Regenerating Cargo.nix on this project https://github.com/psionic-k/cargo2nix-features can test the feature logic.  Lots of context on #211  Seems a like fix is just around the corner.

overlay/make-package-set/full.nix
```diff
@@ -11,7 +11,7 @@
 {
   packageFun,
   workspaceSrc ? null,
-  rustChannel,
+  rustToolchain,
```
mainrs, line 14: Maybe I am not understanding the change correctly, but shouldn't this work?

```nix
{
  description = "Workspace flake";

  inputs = {
    cargo2nix.url = "github:cargo2nix/cargo2nix?ref=be-friendly-to-users";
    flake-compat = {
      url = "github:edolstra/flake-compat";
      flake = false;
    };
    flake-utils.url = "github:numtide/flake-utils";
    nixpkgs.url = "github:nixos/nixpkgs?ref=release-21.05";
    rust-overlay.url = "github:oxalica/rust-overlay";
    rust-overlay.inputs.nixpkgs.follows = "nixpkgs";
    rust-overlay.inputs.flake-utils.follows = "flake-utils";
  };

  outputs = { self, nixpkgs, cargo2nix, flake-utils, rust-overlay, ... }:
    flake-utils.lib.eachDefaultSystem (system:
      let
        overlays = [ (import "${cargo2nix}/overlay") rust-overlay.overlay ];
        pkgs = import nixpkgs {
          inherit overlays system;
        };

          rustToolchain= "nightly-2021-12-05";
        };

        workspaceShell = rustPkgs.workspaceShell {};

      in
      rec {
        devShell = workspaceShell;
      }
    );
}
```

psionic-k: `full.nix` is only called by `simplified.nix`.   Sorry for the confusing arrangement.  The `rustToolchain` here is an internal result.  When you call `makePackageSet`` it calls into `simplified.nix` first.

You want to look at the changes to simplified.nix to understand the new logic.  The argument documentation was also updated in the root flake.nix.

passing `rustToolchain` completely overrides the toolchain.  Cargo2nix will just trust that whatever you gave it in `rustToolchain` contains rustc, cargo, std, and rust-src components at the usual paths.  This is full user control with no hand holding.

The normal user will now just pass `rustVersion` and get a `stable.${rustVersion}.minimal...` toolchain from rust-overlay

For nightly, you want to pass `rustChannel = "nightly"` and `rustVersion=YYYY-MM-DD`.  Search for `selectLatestNightlyWith` to see exactly how this gets handled after the changes.

Finally, I've added a combined overlay, so you can completely omit the `rust-overlay` inputs in your example and just instantiate nixpkgs with `cargo2nix.overlay`

mainrs: Thanks for the clarifications!

psionic-k:
Update to features logic is at a rough draft level.  I think it does way more resolutions than are necessary, making it pretty slow, about 30-60s to nixify RustAnalyzer.  Probably can re-use a lot of resolutions.  The big-O complexity is not good * not good (cargo already says it's not good before we run it a bunch of times). 

There's a simplification I didn't put back in yet where if an optional dependency is enabled by a feature that enables its dependent (identical activation), we can skip the logic expression on the optional dep.

One change I've made is that top level crates can have their default features turned off but cannot be entirely turned off.  If a crate activates a feature on a dependency, the only way to expose that to cargo2nix is to put that feature into a top level crate feature.  `default = ["foo/bar"]` works.  I believe this is good because it actually makes all of the dependent crates get built with identical features more often, which is good for caching.  Features are by design not supposed to hurt anything when not needed by a dependent.

Going to clean up this PR tomorrow and add in workspace offsets so that child crates that try to read resources from the workspace level will work.  I'll get the performance to something acceptable and add a progress bar lol.  Also going to try to get some of the updates from the other open branch with the failing OSX build.

Auto-links is cool... but I think I'm going to push it because the existing workarounds are good enough and so many crates these days are just vendoring their C deps and requiring the `cc` crate, which already brings along our powerfubal `stdenv` of doom even on Darwin.

Also going to stamp this as the 0.11.0 release :confetti_ball: :partying_face:

psionic-k:
Help is wanted to update the serialization deps.  Darwin is failing to build cargo2nix with or without this PR if serialization is updated.  At a minimum, some version roulette would be appreciated.

psionic-k:
Found and fixed a case of non-deterministic ordering of the platforms.  Inserts now done into a `BTreeSet` to maintain ordering.

psionic-k:
Recent Darwin failures feel like binary incompatibility and non-deterministic output.  CI succeeded.  Then it did not.  Same issue can migrate between specific steps, so sometimes cargo2nix fails, other time RustAnalyzer fails.

psionic-k:
```
ld: warning: duplicate symbol '_git_buf_puts' in:\n              /nix/store/w7zwa5c0khv42labx1pfg8phisx5m0x7-crate-libgit2-sys-0.12.26+1.3.0/build_script_output/build/libgit2.a(buffer.o)\n              /nix/store/kymlakdqy9yqf2izj74vjv3cfd4mm7vc-crate-git2-curl-0.14.1/lib/libgit2_curl.rlib(buffer.o)\n 
```
Looks like crate libgit2 is trying to build what libgit2_curl already has.    This is not usually super hard to fix.  Adding a step to the buildPhase will likely fix it right up.  Don't know why it's not consistent though.

psionic-k:
I won't be working on this PR for about a week.  Please work on the library compatibility on Darwin.  I can provide more details but won't be on my mac any sooner than a week.

templates/Cargo.nix.tera
```diff
@@ -67,7 +67,8 @@ in
     {%- if crate.source.Local.path == "." %}
     src = fetchCrateLocal workspaceSrc;
     {%- else %}
-    src = fetchCrateLocal (workspaceSrc + "/{{ crate.source.Local.path }}");
+    src = fetchCrateLocal workspaceSrc;
```
sphw, line 70: This is causing issues for me, with a mono repo style setup. Our structure looks something like this
```
├── services
│   └── test-service
│       ├── Cargo.toml
│       └── src
│           └── main.rs
└── shared
    └── rust
        ├── Cargo.toml
        ├── bar
        │   ├── Cargo.toml
        │   └── src
        │       └── lib.rs
        └── foo
            ├── Cargo.toml
            └── src
                └── lib.rs
```

We might want to reference `foo` from `test-service`, which would be done by setting the crate path to `../../shared/rust/foo`. With these changes that no longer works, since `workspaceSrc` is `./.`, and so Nix does't know about the relative path.

It seems like the appropriate fix would be to somehow get the per-crate root package, and use that in place of `workspaceSrc`. You would also need to change `workspaceOffset` to be aware of the offset as well. I am not familiar enough with Cargo to know if that info is easily attainable though.

psionic-k: Just so I'm clear what needs supporting, there is no top level Cargo.toml right?  shared and services are just directories of a monorepo?

I'm somewhat confused how the previous setup works because the Cargo.nix `packageFun` argument is at most for one workspace right now.  What does the expression look like where `../../shared/rust/foo` is a dependency?

psionic-k: The diff for the Cargo.nix would be helpful.  I guess there's more information in `crate.source.Local.path` than I understood.

psionic-k: Okay I see several ways that this commit needs improvement

The basic strategy is to build with the full source available but to move any workspace toml out of the way since it will interfere with the synthetic toml created just before building the isolated crate.

Since the relative paths can point quite far over in the source, the workspace Cargo.toml needs to be disabled after the directory change (in mkcrate.nix).

I don't think I tested this well enough.  Thanks for using the branch and catching this issue :+1:

sphw: > Just so I'm clear what needs supporting, there is no top level Cargo.toml right? shared and services are just directories of a monorepo?

In this instance `shared/rust` is a workspace, but I don't think that has a big impact on the issue. `services` is not a workspace. But the individual services within the `services` folder, are often workspaces. 

> The diff for the Cargo.nix would be helpful. I guess there's more information in crate.source.Local.path than I understood.

Here is an example of the diff. 
```
<     src = fetchCrateLocal workspaceSrc;
<     workspaceOffset = "../../../../shared/rust/fortanix-enclaves";
---
>     src = fetchCrateLocal (workspaceSrc + "/../../../../shared/rust/fortanix-enclaves");
```
 For my `Cargo.nix`files `workspaceSrc` is `./`, and so the relative path was just fetched by Nix like any other path. It seems like cargo2nix isn't aware of the workspace for `shared/rust`, which in my use case is fine since we don't reference any files outside of an individual crate.

psionic-k: I had to revert the commit.

The cargo library behavior I tried out is such that I would have to re-implement the workspace discovery to find the workspace for each crate to use it when setting the source.

I was able to make some workarounds for similar situations using the offset.  The idea might have a future for such cases.

A potential wrinkle I uncovered while exploring was that we could have .cargo directories declared in parent directories, and these could interfere with settings while building each crate.

psionic-k:
On vacation, but did a quick pass with 1.58.1 and newer nixpkgs and got same issue on Darwin with duplicate symbols.

psionic-k:
Checking on different branches of nixpkgs to try to help out darwin. 

Nothing works when shelled into the build.  No leads so far.  libgit2-sys is building git2 via a build script.  I don't see another version anywhere, so possibly clang (noticed it's doing the linking) might be screwing up.

I'm doing a bit of version roulette and then going to look at the situation in libgit2-sys and see how we're getting libgit2 built and try to understand why that fails to link later when building the cargo crate.

alarmak:
@psionic-k 
_It's truly astounding how many different ways you find to waste our time_

After correcting your spelling of the `propagatedNativeBuildInputs` in the libgit2-sys crate override, I have corrected your folly so that we may proceed at last.  I expected more of you.

```shell
./result-bin/bin/cargo2nix --version
0.11.0
uname -s
darwin
```

You're welcome

overlay/overrides.nix
```diff
@@ -149,11 +149,11 @@ in rec {
       name = "libgit2-sys";
       overrideAttrs = drv: {
         propagatedBuildInputs = drv.propagatedBuildInputs or [ ] ++ [
-          pkgs.libgit2
```
alarmak, line 152: Such brilliant reasoning @psionic-k

psionic-k:
Added some drive by fixes from ideas in #228 .  README overhaul close to final.  Getting ready to tag & release this as 0.11.0

PR #236, opened 2022-03-17
Add compatibility for projects without nix flakes

superboum:
We are developing [garage](https://git.deuxfleurs.fr/Deuxfleurs/garage) and we want to use only stable nix features. As a result, we want to avoid Nix flakes as long as they are flagged as "experimental". 

We added this small wrapper on our forked repository to be able to import cargo2nix from our main project. 

I think this patch could benefit other users by being integrated upstream :smiley:

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,10 @@
+(import
+  (
+    let lock = builtins.fromJSON (builtins.readFile ./flake.lock); in
+    fetchTarball {
+      url = "https://github.com/edolstra/flake-compat/archive/${lock.nodes.flake-compat.locked.rev}.tar.gz";
+      sha256 = lock.nodes.flake-compat.locked.narHash;
+    }
+  )
+  { src = ./.; }
+).defaultNix
--- a/flake.lock
+++ b/flake.lock
@@ -1,5 +1,21 @@
 {
   "nodes": {
+    "flake-compat": {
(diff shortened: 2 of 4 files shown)


superboum:
I am not sure to understand your previous message, especially this part: "This support would be going full circle on flake compat FYI. It's been there and not been there". 

Maybe I am not aware that you already discussed the possibility to make cargo2nix retro-compatible and dismissed it for a specific reason?

psionic-k:
@superboum   IIRC I supported flake compat right after switching to cargo2nix to flakes.

I made a judgement that **most people will be better off in the flakes world than persisting with other solutions to pin inputs** etc.  This could be wrong in some cases.  The reason I say FYI is to emphasize the warning / advice to switch to flakes.

However, last time I used flakes on Mac, I ran into some annoying issues that required hacky workarounds.  For that reason, I can believe that "experimental" is not just a label and can support adding back in the compat support for a bit longer.

It might have been compat that makes `nix flake show` a little bit less pretty, throwing some warnings.  Issues like this are minor though.

psionic-k:
I'll review this back on my dev machine and likely merge it in.   Thanks for changes.

psionic-k:
Closing for #224 

Changes were added and attributed.  Thanks

PR #237, opened 2022-03-17
openssl-sys expects a rust triple

superboum:
`openssl-sys` expects a rust triple and not a nix triple.
The build break when the rust and the nix triple are different, for example on armv6.
It is the same kind of bugs as the one I fixed on #201.

This patch has been tested on our project where we use cargo2nix, openssl & cross compile to armv6:
  - this is what a failed build looks like, before the patch: [#1391](https://drone.deuxfleurs.fr/Deuxfleurs/garage/1391/5/3)
  - this is what a working build looks like, after the patch: [#1408](https://drone.deuxfleurs.fr/Deuxfleurs/garage/1408/5/3)


psionic-k:
Looks correct on inspection.  Merging after CI.  Thanks for changes!

psionic-k (approved):

PR #242, opened 2022-05-02
Add support for nix 2.8

DarkKirb:
Nix 2.8 broke defaultApp, which has been replaced by apps.default. If you tried running cargo2nix on nix 2.8 currently you would get the following error message:

```
error: attribute 'defaultApp.x86_64-linux' should have type 'derivation'
```


psionic-k:
Confirmed and appended changes to #224 (with your name on changes)

psionic-k:
Confirmed and appended changes to #224 (with your name on changes)

PR #260, opened 2022-05-21
Remove extraRustComponents from extraArgs

alias-dev:
Resolves an issue where `makePackageSetInternal` is called with an unexpected argument when `extraRustComponents` is set.


psionic-k (approved):
Appreciate changes.  It is a good day.  Merging after green.

PR #274, opened 2022-06-13
use a more robust method to find crates in multi-crate workspaces

jmpunkt:
### Motivation
Currently cargo2nix does not build any project which uses diesel. The method used to locate crates within a multi-crate workspace matches multiple crates. In this particular case, [grep](https://github.com/cargo2nix/cargo2nix/blob/5127b95a61bbf211f975ea9faf2fb31251a4a413/overlay/mkcrate.nix#L162) matches the crates diesel and diesel_cli. Therefore, the following `cd` command fails, since two paths are specified.

### Change
Instead of using a grep, use cargo metadata and parse the output with jq.

### DIFF
--- a/overlay/mkcrate-utils.sh
+++ b/overlay/mkcrate-utils.sh
@@ -296,3 +296,11 @@ cargoVerbosityLevel() {
 
   echo ${verbose_flag}
 }
+
+cargoRelativeManifest() {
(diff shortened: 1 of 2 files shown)


psionic-k (approved):
Looks like I could have just written a more specific regex, but the structured solutions are preferable as long as the binaries work consistently in the sandbox.

Thanks for changes!  Merging.

PR #276, opened 2022-06-14
Run pre/post install hooks by default

### LINKED ISSUE #275: Standard install hooks aren't run by default
"Standard" according to the [nixpkgs manual](https://nixos.org/manual/nixpkgs/stable/#ssec-install-phase). In particular I was trying to add a `postInstall = "mv $bin/bin/something $bin/bin/something-else";` to my package but it didn't work. It took me a while but eventually I found that `cargo2nix`'s `installPhase` doesn't run any of the hooks: https://github.com/cargo2nix/cargo2nix/blob/b57cfe7da3d34274c1db1f4273d42d016147d254/overlay/mkcrate.nix#L317-L330

I found this behavior to be somewhat surprising. I'm not sure whether this was intentional, but I thought I'd point it out.

CobaltCause:
This is more consistent with the [nixpkgs manual][0]. Code adapted from [here in nixpkgs][1].

[0]: https://nixos.org/manual/nixpkgs/stable/#ssec-install-phase

[1]: https://github.com/NixOS/nixpkgs/blob/d8916eaeaf7080aa0642122d34d04f683ccf1f17/pkgs/stdenv/generic/setup.sh#L1134-L1155

Not sure if there are tests somewhere I need to update in this repository, though I did [test with my own code](https://or.computer.surgery/charles/cargo-shim/-/commit/1b4051ca1d698fcdd43fda674b500d01e3f43b63) that I originally wanted this feature for; you can try it yourself with `nix run 'git+https://or.computer.surgery/charles/cargo-shim?ref=cargo2nix-install-hooks'` (it just prints a help message then exits). You can tell it works because if it didn't, the rename would never happen and thus `nix run` would complain that it can't find the file specified in `apps.default.program`.

Fixes #275.

### DIFF
--- a/overlay/mkcrate.nix
+++ b/overlay/mkcrate.nix
@@ -314,7 +314,9 @@ let
 
     outputs = ["out" "bin"];
 
-    installPhase = if compileMode != "doctest" then ''
+    installPhase = ''
+      runHook preInstall
+    '' + (if compileMode != "doctest" then ''
       mkdir -p $out/lib
       cargo_links="$(remarshal -if toml -of json Cargo.original.toml | jq -r '.package.links | select(. != null)')"
       if (( MINOR_RUSTC_VERSION < 41 )); then
@@ -327,6 +329,8 @@ let
       mkdir -p $bin
       touch results.json
       mv results.json $out/share/
+    '') + ''
+      runHook postInstall
     '';
   };
 in


psionic-k:
Looks pretty straightforward.  I'll follow up with the remaining hooks since they are also very straightforward to support.

psionic-k (approved):

PR #284, opened 2022-07-07
Fix `zsh: no matches found: github:cargo2nix/cargo2nix#bootstrap`

yvan-sraka:


yvan-sraka:
I just discover that's not default `zsh` who complain when argument contains `:` not enclosed in quotes, that's the GRML config I'm using …

Issue #232, opened 2022-01-04
Is it possible to change the linker used?


mainrs:
I wanted to try [mold](https://github.com/rui314/mold) out. Is it possible to configure the linker without having to fiddle around with `.cargo/config`? The latter isn't really possible, since the specified linker has to be mentioned using its absolute path. And getting that path if the linker is installed via a flake seems impossible.

psionic-k:
It's not hard, but you will need to hack something in.  To start with, I would go ahead and check out a copy of cargo2nix.

Each crate is built independently by invoking cargo in such a way that it works in a sandbox.  To configure this, the original Cargo.toml is replaced with a synthetic one and a .cargo/config is created in place.  This happens inside the `configureCargo` phase of mkcrate.nix https://github.com/cargo2nix/cargo2nix/blob/master/overlay/mkcrate.nix#L161-L189

I think you just want to adjust how the .cargo/config is set up or if cargo can accept a linker argument, you can write in some hacks to configure it that way. 

Recommend hacking first and worrying about code longevity after there's a working example available

mainrs:
Thanks for guiding me through it! I'll make sure to update the issue once I have more time. Might be a week or two!

psionic-k:
If I was going to beat you to it, I would need some prodding about specific linker setups to try and how they are usually done

mainrs:
> If I was going to beat you to it, I would need some prodding about specific linker setups to try and how they are usually done

We can maintain a list of linkers and what needs to be taken care of for each of them. I only really read into `mold` but here is the stuff I noticed:

### [mold](https://github.com/rui314/mold#how-to-use)

- Change linker using `.cargo/config`. Needs absolute path into the nix store.
- `gcc` doesn't support an [arbitrary path to -fuse-ld](https://stackoverflow.com/a/67645678). It's recommended to use `mold` with `clang`.
- Compilation had problems and didn't find `GLIBC_2.33'. I think (?) this is due to not using `gcc`. I tried adding `pkgs.stdenv.cc` to my `buildInputs`, that didn't work though.

psionic-k:
> Compilation had problems and didn't find GLIBC_2.33. I think (?) this is due to not using gcc. I tried adding pkgs.stdenv.cc to my buildInputs, that didn't work though.

Which compilation?

It's definitely just a bit of work on mkcrate and then some configuration to the nixpkgs configuration to swap out linkers.  You might see if they have a mold configuration to look at in system & crossSystem values.

[...]

PR #289, opened 2022-08-11
Fix cross compiling for Fortanix SGX projects

sphw:
This PR fixes two issues when cross compiling SGX projects. The first is that sgx requires a particular version of`ld`, similar to wasm-ld. The second is that the `sgx` triple does not get parsed properly, so a manual override needs to be mad

### DIFF
--- a/overlay/mkcrate.nix
+++ b/overlay/mkcrate.nix
@@ -184,7 +184,8 @@ let
       ]
 
     # HACK: 2019-08-01: wasm32-wasi always uses `wasm-ld`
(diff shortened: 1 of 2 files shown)


psionic-k:
Changes look good to me.  Thanks.

psionic-k:
Cherry-picked into #292 

Thanks for changes.  Looking more obvious about what linker options will be needed for a more generic treatment.

#232 

Closing in favor of #292 

Changes in that branch will soon appear in `unstable` and then next release, 0.12.0.  After merging in fixes from forks.

Issue #238, opened 2022-03-22
`workspaceShell`: "Argument list too long"


mainrs:
When using `workspaceShell` I get this error when running `nix develop` on the flake below:

```console
$ nix develop
error: builder for '/nix/store/hnj8smywwcvynfrvll603ssmzzn99lzw-nix-shell-env.drv' failed with exit code 1;
       last 2 log lines:
       >
       > ErroSysError{executing '/nix/store/4nmqxajzaf60yjribkgvj5j54x9yvr1r-bash-5.1-p12/bin/bash': Argument list too long
       For full logs, run 'nix log /nix/store/hnj8smywwcvynfrvll603ssmzzn99lzw-nix-shell-env.drv'.

$ nix log /nix/store/hnj8smywwcvynfrvll603ssmzzn99lzw-nix-shell-env.drv
^A
^E^@^@^@^@^@^@^@Error^@^@^@^@^@^@^@^@^@^@^@^H^@^@^@^@^@^@^@SysError{^@^@^@^@^@^@^@executing '/nix/store/4nmqxajzaf60yjribkgvj5j54x9yvr1r-bash-5.1-p12/bin/bash': Argument list too long^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@
```

```nix
{
  inputs = {
    cargo2nix.url = "github:cargo2nix/cargo2nix";
    flake-compat = {
      url = "github:edolstra/flake-compat";
      flake = false;
    };
    flake-utils.url = "github:numtide/flake-utils";
    nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
    rust-overlay.url = "github:oxalica/rust-overlay";
    rust-overlay.inputs.nixpkgs.follows = "nixpkgs";
    rust-overlay.inputs.flake-utils.follows = "flake-utils";
  };
  outputs = { self, cargo2nix, flake-utils, nixpkgs, rust-overlay, ... }:
    flake-utils.lib.eachDefaultSystem (system: 
      let 
        overlays = [
          (import "${cargo2nix}/overlay")
          rust-overlay.overlay
        ];
        pkgs = import nixpkgs {
          inherit overlays system;
        };

        rustPkgs = pkgs.rustBuilder.makePackageSet' {
          rustChannel = "1.59.0";
          packageFun = import ./Cargo.nix;
        };

        workspaceShell = rustPkgs.workspaceShell {};
      in
      {
[...]

PR #290, opened 2022-08-11
Deduplicate propagatedBuildInputs by store path

sconybeare:
This is intended to fix https://github.com/cargo2nix/cargo2nix/issues/238. It works by using the nix store path to deduplicate the propagatedBuildInputs of each crate. Previously, a dense dependency graph could result in an exponentially large number of duplicate dependencies on packages like openssl or pkg-config. With this change, these dependencies get deduplicated and the exponential blowup is avoided.

### DIFF
--- a/overlay/mkcrate.nix
+++ b/overlay/mkcrate.nix
@@ -122,7 +122,13 @@ let
     name = "crate-${name}-${version}${optionalString (compileMode != "build") "-${compileMode}"}";
 
     buildInputs = runtimeDependencies ++ lib.optionals stdenv.hostPlatform.isDarwin [ pkgs.libiconv ];
-    propagatedBuildInputs = (concatMap (drv: drv.propagatedBuildInputs) runtimeDependencies);
+    propagatedBuildInputs =
+      let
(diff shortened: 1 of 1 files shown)


sconybeare:
@psionic-k sorry about the drive-by patch -- I see in the thread for #238 that you have a more comprehensive fix in mind, but I think this will work pretty well as a stopgap measure in the meantime.

sconybeare:
Err, this apparently doesn't actually fix the bug with workspaceShell -- it *does* fix a related bug where `nix build` fails with an "Argument list too long" error.

psionic-k:
If deps are being amplified through multiple inclusion (DAG structure) and multiple levels, the right place to handle some of the problem is at each derivation when being created.  I will likely include this.

The proper solution for the too many arguments is using `_passthru` on overrides that need to set a shell variable in the user's development shell.  Should be short work.

Thanks for changes.  Working on getting them into unstable right now

psionic-k:
I've included a slightly more compact fix in #292 

https://github.com/cargo2nix/cargo2nix/pull/292/commits/10f8429e455495970b8f94031e7a57e530bdfdf1

Closing in favor of #292 

Will punch `unstable` with the results of that branch after I finish gathering up fixes from the forks

PR #292, opened 2022-09-27
Unstable, next release CI

psionic-k:
Hitting CI before I punch the unstable branch.  History snapshot of where current work picks up from open PR's

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v3
-      - uses: cachix/install-nix-action@v20
+      - uses: cachix/install-nix-action@v22
         with:
           extra_nix_config: |
             log-lines = 200
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,13 +25,12 @@ jobs:
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-      - uses: actions/checkout@v2.4.2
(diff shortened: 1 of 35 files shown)


psionic-k:
Just squashed some commits I was throwing at CI while my macbook is charging up.  The apple frameworks seem to require propagation still.  I'm unclear if this is because of a difference in setupHook & linking propagation or because the dependencies are actually different in a way that requires dependents to also depend on them.

Anyway, workspace shell seems to be in a much better place.  TODO's were noted.

Dev dependencies are implicitly handled in the workspace shell because it looks at the dependencies for every single crate in the Cargo.nix, meaning it won't miss dev dependencies.

psionic-k:
I just force-pushed with a preference for `shellHook` instead of `setupHook` for injecting environment into derivations and propagating that injection to the workspace shell.

I've completely removed the derivation hack in propagateEnv that was part of the mechanism of generating huge lists of propagated inputs for no good reason.  This may break some overrides without visibility due to the lack of coverage on pq-sys and some other crates.

The current `show-derivation` result on the shell:

"shellHook": "export PGK_CONFIG_ALLOW_CROSS=1\n\nexport X86_64_UNKNOWN_LINUX_GNU_OPENSSL_DIR='/nix/store/y36xbx84k3640x3l0g66ghqlymidpf6f-openssl'\nexport X86_64_UNKNOWN_LINUX_GNU_OPENSSL_DIR='/nix/store/y36xbx84k3640x3l0g66ghqlymidpf6f-openssl'\nexport OPENSSL_NO_VENDOR=1 # fixed 0.9.60\nexport RUSTFLAGS=\"${RUSTFLAGS:-} --cfg ossl111 --cfg ossl110 --cfg ossl101\"\n",

This is what we want, one inclusion of each dependency's shell setup.  I am still slightly unsure of the distinction with `setupHook` and why we would prefer one or the other or how they should fit into an overall scheme.

PR #296, opened 2022-10-03
Fix DEP_* values sanitization when whitespace among them

litchipi:
Context: When compiling GDK3, the DEP_GTK3-BACKENDS value is `broadway wayland x11` but as it's not correctly quoted, the resulting command is:
`env CC=<stuff> < ... > DEP_GTK3-BACKENDS=broadway wayland x11`
Resulting into a:
`env: wayland not found`
This quick and dirty fix allows to quote the value properly.
I didn't find any clever bash trick to make the code prettier than this ugly fix, advices welcome on this :-)

### DIFF
--- a/overlay/mkcrate.nix
+++ b/overlay/mkcrate.nix
@@ -305,13 +305,23 @@ let
         if (( NIX_DEBUG >= 1 )); then
           set -x
         fi
-        env \
-          "CC_${stdenv.buildPlatform.config}"="${ccForBuild}" \
-          "CXX_${stdenv.buildPlatform.config}"="${cxxForBuild}" \
-          "CC_${rustHostTriple}"="${ccForHost}" \
-          "CXX_${rustHostTriple}"="${cxxForHost}" \
-          "''${depKeys[@]}" \
-          ${buildCmd}
+        dep_keys_sanitized="$(echo ''${depKeys[@]})"
+        if [ -z "$dep_keys_sanitized" ]; then
+          env \
+            "CC_${stdenv.buildPlatform.config}"="${ccForBuild}" \
+            "CXX_${stdenv.buildPlatform.config}"="${cxxForBuild}" \
+            "CC_${rustHostTriple}"="${ccForHost}" \
+            "CXX_${rustHostTriple}"="${cxxForHost}" \
+            ${buildCmd}
+        else
+          env \
+            "CC_${stdenv.buildPlatform.config}"="${ccForBuild}" \
+            "CXX_${stdenv.buildPlatform.config}"="${cxxForBuild}" \
+            "CC_${rustHostTriple}"="${ccForHost}" \
+            "CXX_${rustHostTriple}"="${cxxForHost}" \
+            "$dep_keys_sanitized" \
+            ${buildCmd}
+        fi
       )
     '';
 
--- a/overlay/mkcrate.nix
+++ b/overlay/mkcrate.nix
@@ -121,6 +121,10 @@ let
     inherit src version meta NIX_DEBUG;
     name = "crate-${name}-${version}${optionalString (compileMode != "build") "-${compileMode}"}";
 
+    # Adding libiconv is a convenience hack.  It really isn't needed by every
+    # derivation and should instead be added / propagated where appropriate, but
+    # until someone decides to investigate the actual dependencies, it remains
+    # here instead of in overrides.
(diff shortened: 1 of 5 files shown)


psionic-k:
I'm going to get out the bash bible before merging a quick & dirty fix, but thanks for diagnosing the issue.  That was a tricky place to fail.

psionic-k:
I'm preparing a different commit for this.  I want to point out that you should be able to perform some cleanup in the preInstall hook or postBuild hook, depending on when the correct information is available.

overlay/mkcrate.nix
```diff
@@ -305,13 +305,23 @@ let
...
           set -x
         fi
-        env \
-          "CC_${stdenv.buildPlatform.config}"="${ccForBuild}" \
-          "CXX_${stdenv.buildPlatform.config}"="${cxxForBuild}" \
-          "CC_${rustHostTriple}"="${ccForHost}" \
-          "CXX_${rustHostTriple}"="${cxxForHost}" \
-          "''${depKeys[@]}" \
-          ${buildCmd}
+        dep_keys_sanitized="$(echo ''${depKeys[@]})"
```
psionic-k, line 308: I don't think this should be necessary.  Can you provide a test crate?  I'm going to hit CI with a commit that works for recording dep values with spaces, but I'm not sure of the playback in the build phase.

psionic-k:
https://github.com/cargo2nix/cargo2nix/pull/292/commits/beca5239f5179a4f9d15a11495760a73a5f3c38c  Try this

overlay/mkcrate.nix
litchipi: The crate that failed were `gdk-sys`, `gdk` and `gdkx11-sys`

litchipi:
Your commit works perfectly ! :+1:

litchipi:
Well bad news I don't know why it worked but it didn't work twice
Apparently you have to use double quotes instead, 
``` diff
 single_quote_whitespaced() {
 if [[ $1 == *[[:space:]]* ]]; then
-    echo "'$1'"
+    echo '"$1"'
 else
     echo $1
 fi
```

Which now makes the name of the function not so good, but it works now :sweat_smile:

psionic-k:
That is actually really odd. 

`echo '"$1"'` will just output `"$1"` literally.

This would only make sense if, at the time the flag is read, `$1` is set and `"$1"` properly interpolates to the value.  It's likely just picking up whatever `$1` is at that point in the builder.  Since reading dep keys is a function, `$1` is likely unset or set to something benign.

You might need to shell into the offending drv and see what values are present and what's going into the actual output.  https://github.com/cargo2nix/cargo2nix/tree/unstable#declarative-build-debugging-shell

It can be faster and more informative than print line debugging these behaviors.

litchipi:
The problem is that `echo "'$1'"` creates an execution line like this:
```
env CC_x86_64-unknown-linux-gnu=<stuff> CXX_x86_64-unknown-linux-gnu=<stuff> CC_x86_64-unknown-linux-gnu=<stuff> CXX_x86_64-unknown-linux-gnu=<stuff> DEP_GDK-3_INCLUDE=<stuff> 'DEP_GDK-3_BACKENDS='\''broadway' wayland 'x11'\''' DEP__INCLUDE=<stuff> <path to>/cargo build --release --target x86_64-unknown-linux-gnu --no-default-features --message-format json-diagnostic-rendered-ansi
```
Which leaves `wayland` out of the quote, thus creating the `env: 'wayland': No such file or directory`

psionic-k:
That makes sense.  I want to be sure and kill this 100% rather than just leaving some more edge cases for the next person.  The problem would seem extremely likely to have a generic solution, consistency in the write-read round trip.

psionic-k:
Right now I'm running with the two commits from @acristoffers:

https://github.com/acristoffers/cargo2nix/commit/75aa46cf71163d923f09cd94e56710db75f42fe2
https://github.com/acristoffers/cargo2nix/commit/34e9a649dc29ffe4becf65dadfdf0baef36a8ac7

I prefer `readarray` since it's AFAIK the more predictable way to go in this case.

Need to circle back and see if https://github.com/cargo2nix/cargo2nix/commit/beca5239f5179a4f9d15a11495760a73a5f3c38cm is still necessary or even made it into maintenance branch.

Thanks for changes.  Closing in favor of ongoing work on maintenance branch.

PR #297, opened 2022-10-03
Fix the openssl crate by adding the `ossl300` config flag

litchipi:
I don't really understand why the config flags were commented out, but I found that this setup works on nixos-unstable as of today.
For reminder, on `unstable` of NixOS, the OpenSSL got bumped to version 3 per default, and the version `1.1.*` are available under the package `openssl_1_1`, which makes the compilation of the crate fail if we have the version 3 around

### DIFF
--- a/overlay/make-package-set/internal.nix
+++ b/overlay/make-package-set/internal.nix
@@ -71,7 +71,7 @@ lib.fix' (self:
 
   in packageFunWith { mkRustCrate = mkRustCrate'; buildRustPackages = buildRustPackages'; } // {
     inherit rustPackages callPackage pkgs rustToolchain noBuild;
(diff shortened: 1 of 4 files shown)


psionic-k:
My mistake.  Thanks for adding the 300 flag.

You only saw the missing 300 because of unstable branch, correct?  I am using 22.05 in development.  I could add an extra nixpkgs version to CI to detect these issues sooner, but I might also put it behind a flag that just warns when it fails.

psionic-k:
Don't know what happened on CI.  I'll let you have a shot at it until I make it back to this issue.

litchipi:
Yes it only causes a problem on `nixos-unstable`, but I thought the `unstable` branch was synced with nixpkgs `unstable`
Well I let you merge it / delay it however you want, but in fine you may need that chance :sweat_smile:

PR #298, opened 2022-10-03
Add crates overrides

litchipi:
Context: I'm trying to build a Tauri application using Nix.
All these crates requires an overlay, and I thought it would be preferable to add them directly there.
I copied the format used in the file, and it works well like this, feel free to edit it as you want.

Does this file require to be sorted alphabetically ?

### DIFF
--- a/Cargo.nix
+++ b/Cargo.nix
@@ -14,6 +14,7 @@ args@{
   target ? null,
   codegenOpts ? null,
   profileOpts ? null,
+  cargoUnstableFlags ? null,
   rustcLinkFlags ? null,
   rustcBuildFlags ? null,
   mkRustCrate,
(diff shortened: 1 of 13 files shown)


psionic-k:
Sure, let's sort alphabetically.  I would take a look at if your dependencies will work as just build inputs.  It's not usually the case that they need to propagate.

pkg-config from nixpkgs will usually be transitively included because of a crate depending on the pkg-config crate.  This is an example where a dependency must be propagated.  For normal libraries, it's usually fine to leave them in `buildInputs`.

Apple frameworks for some reason still require propagation, perhaps a linking style issue.

You can continue committing on this branch and I will cherry pick over your commits and handle manual conflict.  Attribution is retained.

psionic-k:
I force pushed the file you are working on in #292

litchipi:
I tried to have the minimal amount of propagated inputs, and this setup should now work !
Also I squashed and rebased over `release-branch-maintenance` so it should be cherry-pickable if you want

psionic-k:
LGTM.  I'll handle any nits if any come up.  Will be in unstable and 0.12 (doesn't exist yet) after close

PR #301, opened 2022-10-18
Fix feature detection

Alexis211:
Just to let you know @psionic-k that feature detection is currently broken in unstable. I had to change a line in `main.rs` to make it work, I'm not sure it's the correct fix but without that the `Cargo.nix` file I got didn't contain all possible feature flags but only those corresponding to default features.

Here is the broken `Cargo.nix` generated by `unstable` : https://git.deuxfleurs.fr/Deuxfleurs/garage/src/commit/7c0c2299349d81bdffe38f901770d501c5ad625f/Cargo.nix

Here is the correct `Cargo.nix` generatead with this patch : https://git.deuxfleurs.fr/Deuxfleurs/garage/src/commit/00cf076412b88ae0c6254fc16b9bbe661c4f01f2/Cargo.nix

The other line in `mkcrate.nix` is just for my personnal convenience so that I can use `clippy-driver` instead of `rustc` so that I  have all of the clippy lints in addition to standard rustc warnings when doing a build in our CI pipeline. See here for how we use it: https://git.deuxfleurs.fr/Deuxfleurs/garage/src/commit/00cf076412b88ae0c6254fc16b9bbe661c4f01f2/nix/compile.nix

I'm just opening this PR to open a discussion on this, not asking that you merge any of this in its current form.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -3,11 +3,11 @@
     "flake-compat": {
       "flake": false,
       "locked": {
-        "lastModified": 1650374568,
-        "narHash": "sha256-Z+s0J8/r907g149rllvwhb4pKi8Wam5ij0st8PwAh+E=",
+        "lastModified": 1668681692,
+        "narHash": "sha256-Ht91NGdewz8IQLtWZ9LCeNXMSXHUss+9COoqu6JLmXU=",
         "owner": "edolstra",
         "repo": "flake-compat",
-        "rev": "b4a34015c698c7793d592d66adbab377907a2be8",
+        "rev": "009399224d5e398d03b22badca40a37ac85412a1",
         "type": "github"
       },
       "original": {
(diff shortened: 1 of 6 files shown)


psionic-k:
Looks interesting.  IIRC it's a definite fact that the current structure of the code is farther away than a quick one-line fix, but looking at that variable name makes me suspicious of my intent.  I realized it wasn't going in the right direction and may have overlooked a more grave error introduced while reaching that conclusion.

I need to diff the two generated expressions and check with all of the example crates having their Cargo.nix regenerated.  I think I will add this to CI as it's not much longer than the rebuild whenever making changes to mkcrate.nix

psionic-k:
Commits included in unstable.  Closing.  Thanks for changes.

PR #308, opened 2022-11-12
Add linker hack for some platforms without a viable C toolchain

alois31:
Bare-metal and UEFI targets do not currently have a viable C toolchain in nixpkgs. Link for those using LLD where possible.

## Pull Requests

Many files are generated.  When you draft your changes, please separate out
commits that affect:

- Cargo.lock
- Cargo.nix
- flake.lock

Keeping these changes isolated in specific commits makes it much easier to pull
in your changes in parallel with other features.  Maintainers may harvest your
changes.  We only guarantee to preserve authorship in the git log (when we
remember to do so).

### Creating pull requests

1. Fork this repository into the personal GitHub account
2. Select the appropriate branch, release-<version> for stable changes, unstable
   for breaking changes
3. Make changes on the personal fork
4. Make a Pull Request against this repository
5. **Allow maintainers to make changes to your pull request** (there's a checkbox)
6. Once the pull request has been approved, you will be thanked and observe your
   changes applied with authorship preserved (if we remember)

### DIFF
--- a/overlay/mkcrate.nix
+++ b/overlay/mkcrate.nix
@@ -59,6 +59,14 @@ let
   cc = stdenv.cc;
   ccForHost="${cc}/bin/${targetPrefix}cc";
   cxxForHost="${cc}/bin/${targetPrefix}c++";
+  specialLinkers = let
+      lld-elf = "${buildPackages.llvmPackages_latest.lld}/bin/ld.lld";
+      lld-pe = "${buildPackages.llvmPackages_latest.lld}/bin/lld-link";
+    in {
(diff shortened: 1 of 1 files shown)


psionic-k:
Holding off on this for now.  There's a lot of request to configure the linker and there's already some special cases that make reading a bit taxing.  I think the time for improving mkcrate is now.  I will see what all you've got while making changes.

First things first, we just need people to be able to put the right values into place in the right situations, then try to automate configuration of the "usual" correct choices with known special cases at a higher level while preserving the capability to override.

PR #314, opened 2023-02-22
Increase dependency limits

DarkKirb:
Currently, there is a hard limit to how many direct and total dependencies a crate can have with cargo2nix, caused by linux argument length limits. Certain large crates with hundreds of total dependencies, like yew, cannot be built due to this.

This pull request changes dependency handling by symlinking the propagatedBuildInputs together in a separate build step. Additionally it instructs nix to pass the direct dependency list in as a file. This results in greatly increased dependency limits.

I have tested the changes and they do increase the dependency limits enough to build yew and crates that depend on it. This patch is applicable as-is against 0.11.0

### DIFF
--- a/overlay/mkcrate-utils.sh
+++ b/overlay/mkcrate-utils.sh
@@ -184,11 +184,11 @@ install_crate2() {
     fi
   fi
 
-  loadExternCrateLinkFlags $dependencies >> "$out/lib/.link-flags"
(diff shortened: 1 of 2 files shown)


psionic-k:
Likely the issue has been resolved.  I have saved several PR's and commits aimed at this issue and will take a look at crates that could trigger this issue before deciding on one of the approaches.

Thanks for changes and description.

PR #323, opened 2023-05-04
github actions: update actions; remove unneeded token

ryantm:
The actions were failing on my PR, so maybe bumping them will help

We don't need the token anymore since the action knows how to use a default token automatically.
https://github.com/cachix/install-nix-action/pull/157/files


psionic-k:
Ended up incidentally clobbering.  Thanks for PR though.

PR #325, opened 2023-05-10
Fix building crates with inherited workspace parameters

### LINKED ISSUE #322: Usage with cargo workspace dependencies/package
I'm trying to use cargo2nix with [cargo's workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html).

I started refactoring common dependencies into the `[workspace.dependencies]` table, now builds fail because cargo2nix copies the crates source without the root manifest file. I [created a repository to reproduce the error](https://github.com/fracek/cargo2nix-workspace-issue). Is there a way to work around this?

```txt
mycrate-0.1.0.drv
@nix { "action": "setPhase", "phase": "unpackPhase" }
unpacking sources
unpacking source archive /nix/store/68x7iz3qxvcl6m8jvzhw06whzmw9hq5p-mycrate
source root is mycrate
@nix { "action": "setPhase", "phase": "patchPhase" }
patching sources
@nix { "action": "setPhase", "phase": "configurePhase" }
configuring
error: failed to parse manifest at `/build/mycrate/Cargo.toml`

Caused by:
  error inheriting `edition` from workspace root manifest's `workspace.package.edition`

Caused by:
  failed to find a workspace root
error: failed to parse manifest at `/build/mycrate/Cargo.toml`

Caused by:
  error inheriting `edition` from workspace root manifest's `workspace.package.edition`

Caused by:
  failed to find a workspace root
/nix/store/pgf87rgkzdbqhhsddpligy8xiwar63w8-stdenv-linux/setup: line 123: [: !=: unary operator expected
@nix { "action": "setPhase", "phase": "buildPhase" }
building
error: failed to parse manifest at `/build/mycrate/Cargo.toml`

Caused by:
  error inheriting `edition` from workspace root manifest's `workspace.package.edition`

Caused by:
  failed to find a workspace root
/nix/store/pgf87rgkzdbqhhsddpligy8xiwar63w8-stdenv-linux/setup: line 136: pop_var_context: head of shell_variab>
```

sphw:
Cargo workspaces members can now inherit the workspaces metadata and dependencies. To support this Cargo needed to be updated to the latest version, but even then there were a couple of issues to solve. Cargo2nix already handles dependencies cleanly, since Cargo.lock files render solve all the dependency relations. Package metadata on the other handle causes issues, because cargo2nix deletes the workspace Cargo.toml. 

Initially I attempted to solve this by not deleting the workspace file, but that causes many problems since Cargo refuses to not resolve dependencies for the entire workspace. Instead, this PR uses `cargo metadata` to resolve the metadata of the package before removing the workspace Cargo.toml. That data is then merged into the final package Cargo.toml. 

 Let me know if this the best approach. I have tested it on a number of crates including `wgpu` and `rust-gpu` (which both heavily use this pattern) and it works well. 

Fixes #322

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -42,17 +42,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "23b62fc65de8e4e7f52534fb52b0f3ed04746ae267519eef2a83941e8085068b"
 
 [[package]]
-name = "atty"
-version = "0.2.14"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
-dependencies = [
- "hermit-abi",
- "libc",
- "winapi",
-]
-
-[[package]]
 name = "autocfg"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -72,6 +61,24 @@ dependencies = [
  "object",
  "rustc-demangle",
 ]
+
+[[package]]
+name = "base16ct"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4c7f02d4ea65f2c1853089ffd8d2787bdbc63de2f0d29dedbcf8ccdfa0ccd4cf"
+
+[[package]]
+name = "base64"
+version = "0.13.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9e1b586273c5702936fe7b7d6896644d8be71e6314cfe09d3167c95f712589e8"
+
+[[package]]
+name = "base64ct"
+version = "1.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8c3c1a368f70d6cf7302d78f8f7093da241fb8e8807c05cc9e51a125895a6d5b"
 
 [[package]]
 name = "bitflags"
@@ -97,7 +104,16 @@ dependencies = [
  "block-padding",
  "byte-tools",
  "byteorder",
- "generic-array",
+ "generic-array 0.12.4",
+]
+
+[[package]]
+name = "block-buffer"
+version = "0.10.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3078c7629b62d3f0439517fa394996acacc5cbc91c5a20d8c658e77abd503a71"
+dependencies = [
+ "generic-array 0.14.7",
 ]
 
 [[package]]
@@ -121,6 +137,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "bumpalo"
+version = "3.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9b1ce199063694f33ffb7dd4e0ee620741495c32833cde5aa08f02a0bf96f0c8"
+
+[[package]]
 name = "byte-tools"
 version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -133,12 +155,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "14c189c53d098945499cdfa7ecc63567cf3886b3332b312a5b4585d8d3a6a610"
 
 [[package]]
-name = "bytes"
-version = "1.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c4872d67bab6358e59559027aa3b9157c53d9358c51423c17554809a8858e0f8"
-
-[[package]]
 name = "bytesize"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -146,18 +162,17 @@ checksum = "6c58ec36aac5066d5ca17df51b3e70279f5670a72102f5752cb7e7c856adfc70"
 
 [[package]]
 name = "cargo"
-version = "0.61.1"
+version = "0.70.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f76f22dfcbc8e5aaa4e150373354723efe22b6b2280805f1fb6b1363005e7bab"
+checksum = "798a2174afcd5e45e75aa8b82a84dc6ab086a7abe4b4ae7991ada37a9fff6dbb"
 dependencies = [
  "anyhow",
- "atty",
+ "base64",
  "bytesize",
  "cargo-platform",
  "cargo-util",
  "clap",
  "crates-io",
- "crossbeam-utils",
  "curl",
  "curl-sys",
  "env_logger",
@@ -168,10 +183,14 @@ dependencies = [
  "git2-curl",
  "glob",
  "hex 0.4.3",
+ "hmac",
  "home",
+ "http-auth",
  "humantime",
  "ignore",
  "im-rc",
+ "indexmap",
+ "is-terminal",
  "itertools",
  "jobserver",
  "lazy_static",
@@ -180,27 +199,31 @@ dependencies = [
  "libgit2-sys",
  "log",
  "memchr",
- "num_cpus",
  "opener",
  "os_info",
- "percent-encoding",
+ "pasetors",
+ "pathdiff",
  "rustc-workspace-hack",
  "rustfix",
  "semver",
  "serde",
+ "serde-value",
  "serde_ignored",
  "serde_json",
+ "sha1",
  "shell-escape",
  "strip-ansi-escapes",
  "tar",
  "tempfile",
  "termcolor",
+ "time",
+ "toml 0.7.3",
  "toml_edit",
  "unicode-width",
  "unicode-xid",
  "url",
  "walkdir",
- "winapi",
+ "windows-sys 0.45.0",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


sveitser:
Want to try this out because but ran into this issue:
```
nix run github:cargo2nix/cargo2nix/a2c89c34a86ee651270b5ec4b4f662cdf14c2b82  
error: builder for '/nix/store/7kyd7s6kh4n5jsgvb4q4lb4v13p3l9xj-crate-time-core-0.1.1.drv' failed with exit code 101;
       last 8 log lines:
       > unpacking sources
       > unpacking source archive /nix/store/817g0bfva714xp2rvf31qcabda5djald-time-core-0.1.1.tar.gz
       > source root is time-core-0.1.1
       > setting SOURCE_DATE_EPOCH to timestamp 1153704088 of file time-core-0.1.1/src/util.rs
       > patching sources
       > configuring
       > building
       > error: package `time-core v0.1.1 (/build/time-core-0.1.1)` cannot be built because it requires rustc 1.65.0 or newer, while the currently active rustc version is 1.61.0
       For full logs, run 'nix log /nix/store/7kyd7s6kh4n5jsgvb4q4lb4v13p3l9xj-crate-time-core-0.1.1.drv'.
error: 1 dependencies of derivation '/nix/store/aj0j364l3vcmv19zn3n46gx3by0v3rk9-crate-time-0.3.21.drv' failed to build
error: 1 dependencies of derivation '/nix/store/lvb0g4qswqa6viqrm3y2ncjcnfrphcsk-crate-cargo-0.70.1.drv' failed to build
error: 1 dependencies of derivation '/nix/store/frv0a7y7wpp6v37qay3s7sv6jcslk6hg-crate-cargo2nix-0.11.0.drv' failed to build
```

psionic-k:
> Instead, this PR uses cargo metadata to resolve the metadata of the package before removing the workspace Cargo.toml. That data is then merged into the final package Cargo.toml.

The only other way to resolve this would be to propagate the information at generation time, baking it into the Nix expressions.  One way is done in bash and the other in Rust.

Eliminating extraneous uniqueness is another consideration.

@sveitser  I don't think your issue is related.  I will rely on the standard of using Rust Analyzer build to smoke out issues I can't see in review.

Nemo157:
I tried using this in <https://github.com/kitsune-soc/kitsune/compare/main...Nemo157:kitsune:cargo2nix> and it fails with:

```
error: builder for '/nix/store/3lb8y5g5pfqvnixk8kwa29328fqdbjvg-crate-post-process-0.0.1-pre.1.drv' failed with exit code 101;
       last 10 log lines:
       > /nix/store/blpvf60m29q02c0lc5fyhim30ma4y1vv-stdenv-linux/setup: line 131: [: !=: unary operator expected
       > building
       > error: failed to parse manifest at `/build/post-process/Cargo.toml`
       >
       > Caused by:
       >   error inheriting `version` from workspace root manifest's `workspace.package.version`
       >
       > Caused by:
       >   failed to find a workspace root
       > /nix/store/blpvf60m29q02c0lc5fyhim30ma4y1vv-stdenv-linux/setup: line 144: pop_var_context: head of shell_variables not a function context
       For full logs, run 'nix-store -l /nix/store/3lb8y5g5pfqvnixk8kwa29328fqdbjvg-crate-post-process-0.0.1-pre.1.drv'.
error: 1 dependencies of derivation '/nix/store/dh9i1jf3zv7j6x4pnj05j73b5k5hk6gn-crate-kitsune-0.0.1-pre.1.drv' failed to build
```

The manifest at `/nix/store/j2klasyfv2q8nidk7csd8di0rwssslxb-source/lib/post-process/Cargo.toml` appears to just be a direct copy of the original with `version.workspace = true` still.

Nemo157:
(Just thinking about the approach a little bit, I wonder if `cargo package` would be useful somehow? That performs the `Cargo.toml` rewrite step for you).

sphw:
@Nemo157 I tried reproducing your example by including `post-process` in the Cargo.toml like below:
```
post-process.git = "https://github.com/kitsune-soc/kitsune.git"
post-process.branch = "main"
```

And it works fine, so I'm not sure its fair to say "The only time I'd see that working is if it's a non-virtual manifest and only the root crate uses .workspace = true settings.'"

Instead I believe the issue is with how cargo2nix packages the local crate you are building. When you use a remote crate, which is what I tested above, cargo2nix downloads the entire repos (or crates) source tree, and then uses the correct subdirectory for the required crate. When you build locally however, it only loads the directory for the crate. 

I think the workaround you posted should be safe, though I would like @psionic-k 's input on this, because it fundamentally works the same way that cargo2nix works when loading a crate from a remote source.

Nemo157:
Ah, I hadn't considered git dependencies, I only looked into how local crates worked.

The major problem with my workaround is that necessitates rebuilding all local crates when anything in the workspace changes. I think what really needs to happen is that `fetchCrateLocal` is the step that takes in the whole workspace source and applies the inherited parameters to build the crate's isolated source. I think that needs to be a content-addressed derivation to actually work though, since the workspace source changes every time.

sphw:
> The major problem with my workaround is that necessitates rebuilding all local crates when anything in the workspace changes. I think what really needs to happen is that fetchCrateLocal is the step that takes in the whole workspace source and applies the inherited parameters to build the crate's isolated source. I think that needs to be a content-addressed derivation to actually work though, since the workspace source changes every time.

One option I had explored early on in these changes, was creating a trimmed version of the source tree that only included the workspace's Cargo.toml file. Editing the workspaces Cargo.toml, to just include the relevant crate, proved to be challenging. So instead I reverted to this more brute force method.

I think one option would be to include all the required metadata in the `Cargo.nix` file, and then use that metadata to recreate the crate's Cargo.toml.

psionic-k:
> I think one option would be to include all the required metadata in the Cargo.nix file, and then use that metadata to recreate the crate's Cargo.toml.

There's a tradeoff

The toml can include things that cargo2nix the tool doesn't know about, and if we forget to propagate something from metadata back into the toml at build time, the generated toml is incomplete compared to the real tomls

There's much less churn on the side of changes in behavior with respect to the workspace changes, and supporting these behaviors by munging the workspace is usually pretty straightforward.

I'll keep this in mind since while doing a deeper dive on metadata for features.

psionic-k:
Looks like I got a possible repro on a RustAnalyzer dependency:

```
error: builder for '/nix/store/nd7hxlwbpx082ikixjz3m6vpa072r1cd-crate-limit-0.0.0.drv' failed with exit code 101;
       last 10 log lines:
       > /nix/store/gv2cl6qvvslz5h15vqd89f1rpvrdg5yc-stdenv-linux/setup: line 133: [: !=: unary operator expected
       > building
       > error: failed to parse manifest at `/build/limit/Cargo.toml`
       >
       > Caused by:
       >   error inheriting `edition` from workspace root manifest's `workspace.package.edition`
       >
       > Caused by:
       >   failed to find a workspace root
       > /nix/store/gv2cl6qvvslz5h15vqd89f1rpvrdg5yc-stdenv-linux/setup: line 146: pop_var_context: head of shell_variables not a func
tion context
       For full logs, run 'nix log /nix/store/nd7hxlwbpx082ikixjz3m6vpa072r1cd-crate-limit-0.0.0.drv'.
```

Agree that we may need to do some extra nix expression generation.

Out of most of the code I've seen so far, I sense a need to determine if future keys are more likely to break child crates than fix issues in a forward compatible way.

psionic-k:
In the past, the entire workspace source was provided and a path offset was used to get to the desired crate.  I like this method better.  Crates may sometimes use workspace level files, and this is the only way for them to have access.

For inherited parameters, there's two ways to go.  Generate ahead of time or munge the workspace toml.

The point of the synthetic toml is to rig the crate to succeed with minimum information.  This does somewhat favor rigging the workspace toml similarly because we're only talking about a very modest number of keys.

Some metadata is [available to the package](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates), and this is necessary to obtain the desired outputs in more cases, so I am in favor of blacklisting only keys that will cause problems, such as dependencies.  As long as network is not triggered, we usually have everything we need.

A weird situation that could be created by using the entire source tree is that we specify a dependency that has been built via nix yet cargo **finds** the dependency via the workspace and decides to rebuild it, occasionally succeeding.

For instance, the ide crate in RustAnalyzer is specified for the crate via inheritance.
```toml
ide-completion.workspace = true
```

However, the correct dependency is already included in the Cargo.nix and is already made available via flags for the crate to build successfully.  As long as Cargo doesn't get too smart and try to infer that it should build another crate in the workspace because it just looks appetizing (unlikely), I think we're fine.

```nix
      ide_completion = (rustPackages."unknown".ide-completion."0.0.0" { inherit profileName; }).out;
```

The dependencies in Cargo.nix look good, which is expected since the resolve will always have access to the whole workspace at nix generation time.

.cargo and other folders that can exist at the workspace level have to be handled properly and not just ignored.

Going to try to get by with `cargo build --package` as much as possible to try and limit the fallout from workspace files affecting child crates.

Also going to cleans the dependencies from both the workspace and child crate toml.  That's the main key that can cause issues inside the sandbox.  A blacklist / key removal strategy should scale into any other keys that cause problems.

psionic-k:
Thanks for bringing this up.  Closing in favor of https://github.com/cargo2nix/cargo2nix/pull/333

The work went mostly as described at the end of this conversation.

The only nuance is that the workspace is munged to become a single-member workspace rather than changing the directory into the member and removing the workspace files.  I like this method because it really feels like the most elegant way to consume sources and comply with people's expectations for building with or without Nix.

The use of `--platform` was unnecessary due to the removal of irrelevant workspace members.

More metadata use could clean up some of the cruft in mkcrate.nix.  The remarshall tool requires about 130MB in the store to run.  Considering another tool that pulls in less dependencies (and might not suffer remarshal's bug on table keys that is also likely cruft at this point).

PR #331, opened 2023-10-04
Try overlays

marcuswhybrow:
Revert

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -15,7 +15,7 @@
   
   outputs = inputs: with inputs;
     let
-      overlays = import ./overlay (rust-overlay.overlays.default);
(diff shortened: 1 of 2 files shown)


marcuswhybrow:
Whoops created PR on wrong repo. Sorry

marcuswhybrow:
> Is it good? 🐱

Yeah. I fixed two warnings I was getting using cargo2nix. 1. That rust-overlay is now using `overlays.default` instead of `overlay`. and 2. that `replaceChars` is not `replaceStrings`.

I fixed both, but running `nix run github:marcuswhybrow/cargo2nix` is broken. If I figure out why I'll submit both PRs.

PR #351, opened 2024-03-18
 Use string interpolation to generate Cargo.lock path

Zabot:
When specifying a custom `workspaceSrc` 
```
pkgs.rustBuilder.makePackageSet {
  ...
  workspaceSrc = nix-filter {
    root = ./.;
    include = [
      "crates"
      ./Cargo.toml
      ./Cargo.lock
    ];
  };
}
```

`nix build` will fail with the following error

```
at /nix/store/yr2fm9jbvcvy4p55b6d1s4s5bx9pnhnz-source/Cargo.nix:40:21:

   39|   workspaceSrc = if args.workspaceSrc == null then ./. else args.workspaceSrc;
   40|   currentLockHash = builtins.hashFile "sha256" (workspaceSrc + /Cargo.lock);
     |                     ^
   41|   lockHashIgnored = if ignoreLockHash

error: access to absolute path '/Cargo.lock' is forbidden in pure eval mode (use '--impure' to override)
```

If you do build with `--impure`, the build later fails at the same place, this time failing to find the directory

```
at /nix/store/x2ysy1n9p12jrfk1s5kaxjkwm9q22kgl-source/Cargo.nix:40:21:

   39|   workspaceSrc = if args.workspaceSrc == null then ./. else args.workspaceSrc;
   40|   currentLockHash = builtins.hashFile "sha256" (workspaceSrc + /Cargo.lock);
     |                     ^
   41|   lockHashIgnored = if ignoreLockHash

error: getting status of '/Cargo.lock': No such file or directory
```

Both of these errors are because nix attempts to resolve the `/Cargo.lock` path used to lookup the lockfile as an absolute path on the file system, and fails. Instead we should use string interpolation to append the `/Cargo.lock` path to whatever path was passed in through the workspaceSrc.

### DIFF
--- a/templates/Cargo.nix.tera
+++ b/templates/Cargo.nix.tera
@@ -28,7 +28,7 @@ args@{
 let
   nixifiedLockHash = "{{ cargo_lock_hash }}";
   workspaceSrc = if args.workspaceSrc == null then ./. else args.workspaceSrc;
-  currentLockHash = builtins.hashFile "sha256" (workspaceSrc + /Cargo.lock);
+  currentLockHash = builtins.hashFile "sha256" "${workspaceSrc}/Cargo.lock";
   lockHashIgnored = if ignoreLockHash
                   then builtins.trace "Ignoring lock hash" ignoreLockHash
                   else ignoreLockHash;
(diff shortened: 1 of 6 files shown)


templates/Cargo.nix.tera
```diff
@@ -28,7 +28,7 @@ args@{
 let
   nixifiedLockHash = "{{ cargo_lock_hash }}";
   workspaceSrc = if args.workspaceSrc == null then ./. else args.workspaceSrc;
-  currentLockHash = builtins.hashFile "sha256" (workspaceSrc + /Cargo.lock);
```
psionic-k, line 31: I think I just reviewed a duplicate change for this, but either way this change will be intergrated.  Thanks.

psionic-k (approved):

PR #352, opened 2024-03-21
Filter workspace source using lib.fileset

cbrewster:
Right now if any file in the workspace directory changes, Nix will rebuild all local crates. This becomes a problem in polyglot monorepos where all Rust packages get rebuilt when anything gets changed even if no Rust code in the project changed.

This uses `lib.fileset` to generate a workspace filter where only workspace members are included in the source for local crates.

I'm not sure if this will cause breakage for users, maybe we should put it behind a flag?

### DIFF
--- a/templates/Cargo.nix.tera
+++ b/templates/Cargo.nix.tera
@@ -27,7 +27,16 @@ args@{
 }:
 let
   nixifiedLockHash = "{{ cargo_lock_hash }}";
-  workspaceSrc = if args.workspaceSrc == null then ./. else args.workspaceSrc;
+  workspaceSrc = lib.fileset.toSource rec {
+    root = if args.workspaceSrc == null then ./. else args.workspaceSrc;
+    fileset = lib.fileset.unions [
+      (root + /Cargo.toml)
+      (root + /Cargo.lock)
+      {%- for crate in workspace_members %}
+      (root + /{{ crate.path }})
+      {%- endfor %}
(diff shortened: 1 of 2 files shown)


templates/Cargo.nix.tera
```diff
@@ -27,7 +27,16 @@ args@{
 }:
 let
   nixifiedLockHash = "{{ cargo_lock_hash }}";
-  workspaceSrc = if args.workspaceSrc == null then ./. else args.workspaceSrc;
+  workspaceSrc = lib.fileset.toSource rec {
+    root = if args.workspaceSrc == null then ./. else args.workspaceSrc;
```
psionic-k, line 31: Filtering of the source should be done at the user interface, when using the overlay.  If it's done in the Cargo.nix, it will be inflexible.

psionic-k (changes requested):
Thanks for changes and creating some awareness.  I think this one needs another approach.

I didn't quite follow what's needed for mono-repos since I tend to just avoid them when using Nix (version cocktails are easier to construct).

We have a tradeoff with workspace crates where we can provide the entire source to the build, which is sometimes necessary, or we can provide just that crate and expose the other crates as dependencies.

Ever since workspaces started having inherited workspace properties, using subsets of inputs sources became trecherous.  In a mono repo where the root isn't a workspace, this is less problematic, but Nix wants to know these things up front.

PR #353, opened 2024-03-29
Add override libsqlite3-sys to all

hrdl-github:
This was forgotten in 546b5b0e7406662d65cc0da9eefb2b0067c379a1.


psionic-k:
I've been doing some work that supports a conclusion that the heavy-handed style of managing cargo in cargo2nix is inelegant to enough important problems that I will be discontinuing maintenance and recommending crane soon. :warning:

psionic-k (approved):

PR #355, opened 2024-03-31
Examples: use GitHub cargo2nix URL, not relative path

proski:
The users are shown one thing and told to do another.

The use of parent directories in flake URLs is unreliable (that's why `get-flake` exists) and should be avoided in examples.

Referring to the parent directory creates a circular dependency in `flake.lock` files. `nix flake lock` keeps changing the sources every time it's run because the `flake.lock` files in the examples affect the checksum of the parent flake, which in turn goes to `flake.lock` in the examples.

With GitHub references, the locks would get out of date only when a new commit is merged to the branch used in the URL.

### DIFF
--- a/examples/1-hello-world/flake.lock
+++ b/examples/1-hello-world/flake.lock
@@ -8,14 +8,18 @@
         "rust-overlay": "rust-overlay"
       },
       "locked": {
-        "lastModified": 0,
-        "narHash": "sha256-ptZO38KE7mFYYxTNyPnPk0dE5L4XznugSg+pHlUY1Pk=",
-        "path": "../../",
-        "type": "path"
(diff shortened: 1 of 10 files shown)


proski:
This needs some care, as we probably don't want to endorse a broken branch. Either the ref branch should be the current branch, or it should be the latest release.

proski:
Also, the examples are used in CI. The CI should be using the current code under test, not a GitHub reference.

psionic-k:
Design considerations are just that local testing and that CI is useful / works.

Version mismatch between test workspace Cargo.nix and the cargo overlay will often cause breakage.

Perhaps the flakes should be overridden via automation and scripts.  At the time of CI, the rev under test is available via git.

PR #356, opened 2024-03-31
Configure cargo to avoid any network access

proski:
This became necessary starting with Rust nightly 2024-03-26. But it has never been a good idea to access network in Nix builds.

Both `--offline` and `net.offline` appeared in Rust 1.36.

https://github.com/rust-lang/cargo/pull/6934

### DIFF
--- a/overlay/mkcrate.nix
+++ b/overlay/mkcrate.nix
@@ -178,6 +178,8 @@ let
     configureCargo = ''
       mkdir -p .cargo
       cat > .cargo/config <<'EOF'
+      [net]
+      offline = true
       [target."${rustBuildTriple}"]
       linker = "${ccForBuild}"
     '' + optionalString (codegenOpts != null && codegenOpts ? "${rustBuildTriple}") (''


proski:
It looked for changes in cargo between 2024-03-25 and 2024-03-26. There are many changes, but nothing looks like an obvious candidate. In any case, I remember seeing errors about network access with cargo2nix and a much older Rust long time ago. It was quite confusing to me, as I didn't know at the time that Nix builds are isolated from the net.

psionic-k:
I'll check it out, but I've been considering deprecating this project and telling people to add features to crane

proski:
To reproduce the issue:

`cd examples/1-hello-world`
`cargo add maitake-sync`
patch `flake.nix`:
```
+    rust-overlay.url = "github:oxalica/rust-overlay/42a168449605950935f15ea546f6f770e5f7f629";
+    cargo2nix.inputs.rust-overlay.follows = "rust-overlay";
...
+          rustVersion = "2024-03-28";
+          rustChannel = "nightly";
```
Update `Cargo.lock`, `flake.lock` and `Cargo.nix`

```
error: builder for '/nix/store/30ipzwmzqd892pq64sw6prxprw65aaya-crate-mycelium-bitfield-0.1.5.drv' failed with exit code 101;
       last 10 log lines:
       > warning: spurious network error (2 tries remaining): [6] Couldn't resolve host name (Could not resolve host: index.crates.io)
       > warning: spurious network error (1 tries remaining): [6] Couldn't resolve host name (Could not resolve host: index.crates.io)
       > error: download of config.json failed
```

2024-03-26 would also fail but with a Cargo error:
```
already borrowed: BorrowMutError
```

2024-03-25 should work.

Apply this PR and 2024-03-28 starts working. 2024-03-25 keeps working, 2024-03-26 remains broken (must be a Cargo bug).

proski:
Re-targeted to the `main` branch.

psionic-k (approved):
This may move the ball forward on something.  It is more correct for the current shape of cargo2nix.

I do think the synthetic cargo.toml as one of the tools in use needs to be avoided when possible in the future.  We're hitting it with dynamite when we should only modify the bare essentials so that it is reactive to the build environment, such as for cross compiling.

PR #357, opened 2024-03-31
Quote `url` value in `fetchCrateGit` output

proski:
The special URL syntax is deprecated:
https://github.com/NixOS/rfcs/blob/master/rfcs/0045-deprecate-url-syntax.md


proski:
Re-targeted to the `main` branch

psionic-k (approved):

PR #358, opened 2024-03-31
Quote `$manifest_path` so it doesn't break comparison if empty

proski:
Something is likely wrong if `$manifest_path` is empty, but at least it won't trigger an unhelpful syntax error in bash.


proski:
Re-targeted to the `main` branch.

psionic-k (approved):

PR #359, opened 2024-04-01
Fix spelling of PKG_CONFIG_ALLOW_CROSS

proski:
Issue introduced in 2ae2f57353232344ceec889d9b2787e207961983

### DIFF
--- a/overlay/overrides.nix
+++ b/overlay/overrides.nix
@@ -225,7 +225,7 @@ in rec {
         pkgs.pkg-config
       ];
       shellHook = drv.shellHook or "" + ''
-        export PGK_CONFIG_ALLOW_CROSS=1
+        export PKG_CONFIG_ALLOW_CROSS=1
       '';
     };
   };


proski:
Sadly, cross-compilation of packages that depend on pkg-config is still broken even with this fix. But the bad name is a distraction and should be fixed.

psionic-k:
I've invited you as a maintainer since I'm having to neglect this repository.

I intend to step away from it because I'm finding that cargo usage is going in a direction where heavy-handed manipulation of cargo is a bad idea, but I don't have or expect to have the individual incentive to move it towards what I think is the correct path.

The idea of generating nix expressions from a cargo plugin is sound.  The use of synthetic Cargo.toml and command line flags I don't think is sustainable.  To become healthy, it probably needs radical changes toward composing target directories and re-implementing the feature logic to basically do what cargo metadata does.

Don't execute supply chain attacks :+1: :police_officer:

proski:
Thank you! I'd like to hear your opinion on #360. I don't have much time either, I just want to get cargo2nix to a state where it keeps working for the projects that are using it already.

proski:
Closing this PR. No changes should be made to the `release-0.11.0` branch, as no further 0.11.x releases are expected. The `main` branch doesn't need to be fixed.

Issue #350, opened 2024-03-12
Can't build gtk project


Luxuride:
Hello, I was trying to create simple project with gtk-rs.
My project consists of hello world application with 
```
[dependencies]
gtk = "0.18.1"
```
in `cargo.toml`

I was using `nix develop github:cargo2nix/cargo2nix#bootstrap`, 
```
$ cargo2nix --version
cargo2nix 0.11.0
```
I copied the flake from examples and changed the project name and cargo2nix address to make it work. That made it work normally on fresh hello world project.

With the gtk dependency I get this when I try `nix develop`
```
$ nix develop
error: builder for '/nix/store/n85pvca4ddp2w5w9b8kxmxy9hxdh3bq6-crate-glib-sys-0.18.1.drv' failed with exit code 101;
       last 10 log lines:
       >   or `pkg install pkg-config`, or `apk add pkgconfig` depending on your distribution.
       >
       >   If you've already installed it, ensure the pkg-config command is one of the
       >   directories in the PATH environment variable.
       >
       >   If you did not expect this build to link to a pre-installed system library,
       >   then check documentation of the glib-sys crate for an option to
       >   build the library from source, or disable features or dependencies
       >   that require pkg-config.
       > /nix/store/wr08yanv2bjrphhi5aai12hf2qz5kvic-stdenv-linux/setup: line 131: pop_var_context: head of shell_variables not a function context
       For full logs, run 'nix log /nix/store/n85pvca4ddp2w5w9b8kxmxy9hxdh3bq6-crate-glib-sys-0.18.1.drv'.
error: 1 dependencies of derivation '/nix/store/l2fv6hi4dr2q198mbyk1a2fb4qfydk1b-crate-atk-sys-0.18.0.drv' failed to build
error: 1 dependencies of derivation '/nix/store/sscvf13175dx4y8zyk5lvsa0yzd0msxy-crate-cairo-sys-rs-0.18.2.drv' failed to build
error: 1 dependencies of derivation '/nix/store/hby9mswa421h2x2gbdr5ssjip94mwffl-crate-gdk-pixbuf-sys-0.18.0.drv' failed to build
error: 1 dependencies of derivation '/nix/store/7ng58flaxq2afz49zc5b8f855j07gvz0-crate-gdk-sys-0.18.0.drv' failed to build
[...]

PR #361, opened 2024-04-03
Overhaul Workspace Shell & Overrides

proski:
A bit of history.  The workspace shell started off as a quick solution to get overrides to express some of their side effects within the user's development shell.  Because it didn't work closely with the actual mechanisms of mkShell, the chosen mechanism, from overrides to dev shell, was use of propagated inputs.  This and some cargo culting had resulted in overrides mimicking other overrides, lots of use of propagatedNativeBuildInputs.

First, workspace shell's overhaul is given the actual package set instead of the noBuild set that is used when running tests.  (the tests themselves may need similar overhaul).  The new shell has one simple job:  Make everything required to build every crate, except rust crates themselves, expressed in environment and available from the store.  This has to be done for the entire rust crate DAG.

Second, once the workspace shell could correctly express dependencies and shell hooks etc without the abuse of propagated inputs, the overrides themselves could be made much more sensible, much more consistent with the rest of Nix.

substractLists works fine for derivation filtering

The issue that lead to using attrSets was using the wrong crate outputs to filter on.  After that was fixed, it was no longer necessary to construct attrSets by keys.  If complexity of a dependency set (n^2 because of lib.unique) becomes too high, the attrSet based implementation should become favored again. This won't really show up to users for n < 1000 where n is the immediate dependencies of crates that are not crates themselves.

Apple frameworks must be propagated?  Missing a setupHook?

### DIFF
--- a/overlay/make-package-set/internal.nix
+++ b/overlay/make-package-set/internal.nix
@@ -72,7 +72,7 @@ lib.fix' (self:
 
   in packageFunWith { mkRustCrate = mkRustCrate'; buildRustPackages = buildRustPackages'; } // {
     inherit rustPackages callPackage pkgs rustToolchain noBuild;
-    workspaceShell = workspaceShell { inherit pkgs noBuild rustToolchain; };
+    workspaceShell = workspaceShell { inherit pkgs rustPackages rustToolchain; };
     mkRustCrate = mkRustCrate';
     buildRustPackages = buildRustPackages';
     __splicedPackages = defaultScope;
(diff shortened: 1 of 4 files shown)


proski:
This commit by @psionic-k was originally added to the `release-0.11.0` branch. However, it has been found to be problematic. A project that depends on the `gtk` crate cannot find `pkg-config` (see #350). Another project can still use `pkg-config`, but not when cross-compiling.

Compared to the original commit, the spelling of `PKG_CONFIG_ALLOW_CROSS` has been fixed. However, it's not sufficient to resolve either problem.

This PR should only be merged when the following conditions are met:

* There are tests for the breakage caused by the initial version of this PR
* The tests pass after this PR is modified

proski:
Replaced "Workplace" with "Workspace". It must have been a typo, as there are no other mentions of "workplace" in the sources.

PR #362, opened 2024-04-03
Refactor of toml munging

proski:
```
*  Use entire workspace, begin refactor of toml munging

   The newer findCrate logic is accurate enough that we don't need to provide the
   source path, although we could re-visit this if some sources don't have a
   workspace in the source root.  Cargo still needs to be able to decide to use
   such a source in any case, or we can't generate the expression.

*  Sanitize workspace & package tomls

   Instead of the existing whitelist style jq statements, a blacklist is produced
   to eliminate keys that cause issues in sandbox builds or will conflict with
   configuration via nix expressions

   The workspace toml is reduced so that it becomes a workspace for just a single
   member, the crate we want to build.

   The crate toml is reduced separately.

   The build happens from the workspace level.

*  remove findCrate stage

   The method of using metadata to check for a workspace before munging the desired
   crate's toml needs to do so before we munge the workspace.

   It made sense to get rid of the extraneous phase, especially now that everything
   is refactored a bit.
```

### DIFF
--- a/overlay/mkcrate-utils.sh
+++ b/overlay/mkcrate-utils.sh
@@ -334,20 +334,38 @@ sanitizeTomlForRemarshal () {
     done;
 }
 
-removeTomlDeps () {
+reducePackageToml () {
+    # This function needs to remove any package keys that conflict with
+    # dependency control or target and profile configuration.
+    # https://doc.rust-lang.org/cargo/reference/manifest.html
     local manifestPatch="$3"
+    local registry="$registry"
+    remarshal -if toml -of json "$1" \
+      | jq 'del(."cargo-features",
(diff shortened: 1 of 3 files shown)


proski:
This set of commits by @psionic-k was originally committed to the `release-0.11.0` branch. However, it was found to be causing errors when compiling Cargo packages that refer to libraries in sibling directories using relative path.

This PR should only be merged when the following conditions are met:

* There are tests for the breakage caused by the initial version of this PR
* The tests pass after this PR is modified

proski:
Added a fix for unquoted bash variable in comparison. It fixes a syntax error, but doesn't fix the behavior of this PR.

Issue #379, opened 2024-11-18
Spurious network error when using rust overlay


bebert64:
I'm trying to use cargo2nix to build a crate from my workspace, and everything is working fine as long as I use the rustc version from nixpkgs. I do have a project that requires rust 1.82.0 though, so I'm trying to add an overlay, but I've tried with both Fenix and oxalica and both generate
`warning: spurious network error (3 tries remaining): [6] Could not resolve hostname (Could not resolve host: index.crates.io)`

My flake looks like:
```
{
  inputs = {
    rust-overlay = {
      url = "github:oxalica/rust-overlay";
    };
    nixpkgs.follows = "rust-overlay/nixpkgs";
    cargo2nix = {
      url = "github:cargo2nix/cargo2nix/";
      inputs.nixpkgs.follows = "nixpkgs";
    };
  };

  outputs =
    { self
    , cargo2nix
    , nixpkgs
    , rust-overlay
    , ...
    }:
    let
      system = "x86_64-linux";
      pkgs = import nixpkgs { inherit system; overlays = [ cargo2nix.overlays.default rust-overlay.overlays.default ]; };
      rustPkgs = pkgs.rustBuilder.makePackageSet {
        rustVersion = "1.82.0";
        packageFun = import ./Cargo.nix;
      };
    in
    {
      my-app = (rustPkgs.workspace.my-app { });
	};
}
```
I've run cargo2nix and it generated my Cargo.nix, which I've added to the repo and commited it.
Trying to run `nix build .#wallpapers-manager` results in correctly building the rust toolchain, but as soon as it tries to build the first crates, I'm getting those warning, and it fails right after.
The logs are :
 ```
Running phase: unpackPhase
@nix { "action": "setPhase", "phase": "unpackPhase" }
unpacking source archive /nix/store/z9jl2bw4whiic1pakqyk4hfc2ydib7sk-imagesize-0.13.0.tar.gz
source root is imagesize-0.13.0
Running phase: patchPhase
@nix { "action": "setPhase", "phase": "patchPhase" }
Running phase: updateAutotoolsGnuConfigScriptsPhase
@nix { "action": "setPhase", "phase": "updateAutotoolsGnuConfigScriptsPhase" }
Running phase: configurePhase
@nix { "action": "setPhase", "phase": "configurePhase" }
Running phase: buildPhase
@nix { "action": "setPhase", "phase": "buildPhase" }
[...]

PR #381, opened 2024-12-08
add note about using newer rust versions in readme

haras-unicorn:
i was frustrated that i had to look through the code to see how to use newer rust versions to find out that i just needed to update the rust-overlay version through my flake so i added the steps to do that in the readme if thats ok

as per #379 and #181 it seems that currently only versions up until verison 1.78 work so maybe i should add that disclaimer as well?

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,9 +197,9 @@ what the current output result is.
 how we store information in dependents and replay them back when
 building dependents.  It is vital for building crates in isolation.
 
-#### Newer rust versions
+## Newer Rust versions
 
-In order to use newer rust versions you can include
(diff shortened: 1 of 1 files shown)


hallettj (approved):

hallettj:
This is very helpful, thanks!

hallettj:
Btw I think that #379 and #181 are fixed on the `main` branch. Note that `main` is not the default branch, so you have to specify it in your inputs like this:

```nix
    cargo2nix = {
      url = "github:cargo2nix/cargo2nix/main";
      inputs.nixpkgs.follows = "nixpkgs";
    };
```

Issue #383, opened 2025-03-01
Does not support Rust edition 2024


maddiemort:
If I try to generate a `Cargo.nix` for a crate that uses edition 2024 (introduced in Rust version `1.85.0`), I receive the following error:

```
thread 'main' panicked at src/main.rs:88:10:
Error generating nix expressions: failed to parse manifest at `<redacted>/Cargo.toml`

Caused by:
    0: failed to parse the `edition` key
    1: this version of Cargo is older than the `2024` edition, and only supports `2015`, `2018`, and `2021` editions.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

If I revert the edition to 2021, I also receive this secondary error that requires manual editing of the lockfile:

```
thread 'main' panicked at src/main.rs:88:10:
Error generating nix expressions: failed to parse lock file at: <redacted>/Cargo.lock

Caused by:
    lock file version 4 requires `-Znext-lockfile-bump`
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I'm using the version of `cargo2nix` from branch `main`, but it looks like this problem is present on other branches too.

hallettj:
Thanks for the report! I believe this is now fixed in `main`. I see that you're already using `main`, but for anyone who comes across this issue: you'll need to specify that branch specifically in your input since the default branch is the last stable release:

```nix
inputs = {
  cargo2nix.url = "github:cargo2nix/cargo2nix/main";
};
```

PR #384, opened 2025-03-03
Cargo 0.86.0

kgtkr:
## Pull Requests
* fix: #383
* support cargo lock version 4
  The latest compiler generates `version = 4`, but cargo2nix was rewriting it to `version = 3`, which has now been fixed.


src/platform.rs
```diff
@@ -72,6 +72,7 @@ fn cfg_to_expr(cfg: &CfgExpr, platform_var: &str) -> BoolExpr {
             ("target_feature", v) => {
                 Single(format!("builtins.elem {:?} {}Features", v, platform_var,))
             }
+            ("target_has_atomic", _) => True, // TODO: This is required to compile cargo, but how can I determine this in Nix?
```
kgtkr, line 75: This should be fixed, but can I get this information using functions from nixpkgs?

hallettj: Hi! I don't know the answer myself. I was just perusing the current state of rust builders and came across this PR. I happened to see a pair of commits from @DarkKirb in the cargo2nix network graph that take a shot at addressing the `target_has_atomic` expression:

https://github.com/DarkKirb/cargo2nix/commit/076cd30f8d5314d3f2228e33d01f90dbda71d064
https://github.com/DarkKirb/cargo2nix/commit/29084d950413d317ce5c5901d5f5fd3510a38974

The second one looks like it fixes an issue in the first one. Those are both from the `master` branch in https://github.com/DarkKirb/cargo2nix

DarkKirb: `rustc -Z unstable-options --print target-spec-json` should output the maximum width of atomic instructions in the field `max-atomic-width`. this would unfortunately require an IFD so maybe this could be precomputed for various targets or read from file in the case a target platform json file is supplied.

the values for default supported targets can be found in this crate: https://github.com/rust-lang/rust/tree/master/compiler/rustc_target/src/spec/targets

the metric i wrote ought to work for many hosted targets.

psionic-k:
I've sent out some more admin invites so that I'm not in the way, but I have no productive capacity to give to this project.  There's at least one other person with maintainer role.

hallettj:
> I've sent out some more admin invites so that I'm not in the way, but I have no productive capacity to give to this project. There's at least one other person with maintainer role.

Lol this is very different from the reception to my PRs on crate2nix - I've seen no responses on those. I'm certainly interested in stepping up for maintenance on one of these two projects.

DarkKirb (changes requested):
I have rebased it to the latest version of cargo2nix: https://github.com/DarkKirb/cargo2nix/tree/cargo-0.86.0, otherwise lgtm!

hallettj:
Ok, a question for the group, especially for @psionic-k  since I'm still getting up to speed. I merged this into `main`, but I have not cherry-picked to `release-0.11.0`. That means that building edition 2024 crates is still not working in the release branch. This is a larger change. Is it time to:
1. cherry-pick this change to `release-0.11.0`?
2. or to tag a new release?

I suppose something to consider is that I'd like to add a PR to support import from derivation in the very-near future which might be nice to have in the next release.

I'm making productive use of the maintainer role I was granted, but that does not authorize me to change the default branch so I would need assistance from an admin to get a release out.

I do have a suggestion which I think might simplify the release process: change the default branch to `main`, and publish to Flakehub. That way anyone who wants to track a major version number could use Flakehub's features to do that. We could still maintain release branches if we want to backport stable changes, and use tags for minor and patch version bumps. That would let people with the maintainer role manage releases, and I think it might reduce the pressure to keep updating the release branch with new features.

psionic-k:
The bitrot in my memory is extreme.  I think a release makes sense just to indicate that development is re-starting

> I'm making productive use of the maintainer role I was granted, but that does not authorize me to change the default branch so I would need assistance from an admin to get a release out.

I can fix this.  My goal is to get out of the way as much as possible.

psionic-k:
I've upgraded @DarkKirb and @hallettj to admins.  Ping me to moderate / advise / arbitrate if something can't be decided.  I recommend biasing towards action.  If changes coming from another maintainer are not uncomfortable, then by default just keep going forward.  If things begin to thrash, then it's time to talk things over.

DarkKirb:
i personally think we should cut a 0.12.0 release soon-ish as there have been significant additions. I’ll create a tracking issue

PR #386, opened 2025-03-07
Add target_has_atomic support

DarkKirb:
As described in #384, cargo the crate now has a hard dependency on the cfg flag `target_has_atomic`. This is not a property that nixpkgs exposes to the developer directly, and as such we essentially have to hardcode which atomics are supported where to avoid imports from derivation.

The way I have implemented it here is extensible to any future cfg flags that may need to be supported that depend on properties from the target spec file.
Other cfg flag implementations could also be rewritten to use this new data (although for space reasons I have elected to not include most of the properties in this file as otherwise the generated json would have been about 600kB in size)

The ability to pass target spec files is also exposed to the user. I believe cargo2nix had some amount of ability to use target spec json files already, but it did not use them for cfg resolution i think.


psionic-k:
Did you get my invite to become a maintainer?  I was going to just let you drive and handle others' PRs.  This project is going to remain stalled without allowing more people to push.

DarkKirb:
Yeah but i wanted others to be able to look at my pr first ^^'

PR #396, opened 2025-04-12
release 0.12.0

hallettj:
@DarkKirb since we discussed creating a release, here's an update with the necessary changes. But let me know if you already have release work in progress.

Updates references to the latest release branch to 0.12, and references to the latest version to 0.12.0. I also included an update in the readme to clarify that it's the `main` branch that has the latest fixes and features, not `unstable`.

Once this is merged I plan to:

- create a tag, `v0.12.0`
- create a release in https://github.com/cargo2nix/cargo2nix/releases - I'll do my best to write a changelog
- create a branch, `release-0.12`
- set the repo default branch to `release-0.12`

There is already a branch called `release-0.12.0`. Instead of using that name I chose to drop the minor version, the trailing `.0`, in the new branch name due to my opinion that release branches should track major versions, and so the branch name should only include the major number. That does leave us with two branches with very similar names which could cause confusion. There's a possibility someone out there has a flake that points to `release-0.12.0`, and their build would break if we delete that branch. Maybe it's worth deleting anyway. It would be nice to be able to redirect `release-0.12.0` to `release-0.12`, but I don't know of a way to do that that would work when pulling a commit by branch name.


hallettj:
All steps are complete - the release is out!

PR #397, opened 2025-05-20
fix alternate registry support

DarkKirb:
Alternate registries got turned into crates-io packages incorrectly, and the syntax for them hasn’t been update for the deprecation of the bare url literals


DarkKirb:
A question is if we should special case the alternate registry at all. the “fetch through alternate registry” code works fine with crates.io packages and due to being fixed output derivations won’t duplicate existing fetchCratesIo source tarballs

Issue #399, opened 2025-06-30
`github:cargo2nix/cargo2nix/v0.12.0` fails on Darwin: (signal: 9, SIGKILL: kill)


donn:
Related:
- https://github.com/rust-lang/cargo/issues/13736
- https://github.com/NixOS/nixpkgs/issues/299606

```
error: builder for '/nix/store/8gz87jnyrzmil9pavmsh90d4a3l665ls-crate-libc-0.2.170.drv' failed with exit code 101;
       last 14 log lines:
       > unpacking sources
       > unpacking source archive /nix/store/hzbd17xga0q9vj28wr2gfsvrwmvbgp0i-libc-0.2.170.tar.gz
       > source root is libc-0.2.170
       > setting SOURCE_DATE_EPOCH to timestamp 1153704088 of file libc-0.2.170/tests/const_fn.rs
       > patching sources
       > updateAutotoolsGnuConfigScriptsPhase
       > configuring
       > building
       >    Compiling libc v0.2.170 (/private/tmp/nix-build-crate-libc-0.2.170.drv-0/libc-0.2.170)
       > error: failed to run custom build command for `libc v0.2.170 (/private/tmp/nix-build-crate-libc-0.2.170.drv-0/libc-0.2.170)`
       >
       > Caused by:
       >   process didn't exit successfully: `/private/tmp/nix-build-crate-libc-0.2.170.drv-0/libc-0.2.170/target/release/build/libc-0bae9c17a08345b5/build-script-build` (signal: 9, SIGKILL: kill)
       > /nix/store/sr62lps3id6rbasca3rp5inhrbkdrj1a-stdenv-darwin/setup: line 140: pop_var_context: head of shell_variables not a function context
       For full logs, run:
         nix log /nix/store/8gz87jnyrzmil9pavmsh90d4a3l665ls-crate-libc-0.2.170.drv
error: 1 dependencies of derivation '/nix/store/4sb1i1ll15w1bqk4v9hp0rhl52p6f32g-crate-backtrace-0.3.74.drv' failed to build
error: 1 dependencies of derivation '/nix/store/apjy8rnzqcq0rwm27zz2sybnyhkndc5h-crate-cargo-0.86.0.drv' failed to build
error (ignored): error: cannot unlink "/private/tmp/nix-build-curl-8.4.0.drv-2/curl-8.4.0": Directory not empty
[...]

PR #405, opened 2025-08-25
Update to nixpkgs 25.05

jacobmichels:
^title. This PR moves from using nixpkgs `23.11` to `25.05`. This fixes the build on macOS. https://github.com/cargo2nix/cargo2nix/issues/399

I also fixed warnings from using some macOS specific packages that no longer have any effect.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -54,16 +54,16 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1705099185,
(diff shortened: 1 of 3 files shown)


jacobmichels:
I did not split out changes to flake.lock into a separate commit, as the changes are minimal. Let me know if this is a problem and I can make changes.

jpds:
I tried bumping this to the new `25.11`,  but there seems to be a missing dependency.

